PR #5, opened 2020-02-19
Dispatch to allgather_packed_range from allgather for dynamic size types

lindsayad:
This works currently by:
- Adding a `is_fixed_type` bool to StandardType classes
- doing SFINAE on Communicator::allgather based on `is_fixed_type`

This also adds default `typename Enable = void` template arguments to
`StandardType` and `Packing`. This is useful for enabling/disabling
specializations based on whether data members are fixed sizes or not.
I'm thinking of an example like `DualNumber<T,D>` where that class
may be fixed-size depending on whether D is fixed-size

### DIFF
--- a/src/parallel/include/timpi/parallel_implementation.h
+++ b/src/parallel/include/timpi/parallel_implementation.h
@@ -2712,7 +2712,8 @@ inline void Communicator::gather(const unsigned int root_id,
 
 
 
-template <typename T, typename A>
+template <typename T, typename A,
+          typename std::enable_if<StandardType<T>::is_fixed_type, int>::type>
 inline void Communicator::allgather(const T & sendval,
                                     std::vector<T,A> & recv) const
 {
@@ -2734,6 +2735,26 @@ inline void Communicator::allgather(const T & sendval,
     recv[0] = sendval;
 }
 
+template <typename T, typename A,
+          typename std::enable_if<!StandardType<T>::is_fixed_type, int>::type>
+inline void Communicator::allgather(const T & sendval,
+                                    std::vector<T,A> & recv) const
+{
+  TIMPI_LOG_SCOPE ("allgather()","Parallel");
+
+  timpi_assert(this->size());
+  recv.resize(this->size());
+
+  unsigned int comm_size = this->size();
+  if (comm_size > 1)
+    {
+      std::vector<T> range = {sendval};
+
+      allgather_packed_range((void *)(NULL), range.begin(), range.end(), recv.begin());
+    }
+  else if (comm_size > 0)
+    recv[0] = sendval;
+}
 
 
 template <typename T, typename A>
--- a/src/parallel/include/timpi/standard_type.h
+++ b/src/parallel/include/timpi/standard_type.h
@@ -56,12 +56,11 @@ struct standardtype_dependent_false : std::false_type
  * More complicated data types may need to provide a pointer-to-T so
  * that we can use MPI_Address without constructing a new T.
  */
-template <typename T>
+template <typename T, typename Enable = void>
 class StandardType : public DataType
 {
-  // Get a slightly better compiler diagnostic
-  static_assert(standardtype_dependent_false<T>::value,
-                "Only specializations of StandardType may be used, did you forget to include a header file (e.g. parallel_algebra.h)?");
+public:
+  static const bool is_fixed_type = false;
 
   /*
    * The unspecialized class is useless, so we make its constructor
@@ -87,6 +86,8 @@ private:
   public:                                                            \
     explicit                                                         \
       StandardType(const cxxtype * = nullptr) : DataType(mpitype) {} \
+                                                                     \
+    static const bool is_fixed_type = true;                          \
   }
 
 #else
@@ -98,6 +99,8 @@ private:
   public:                                                     \
     explicit                                                  \
       StandardType(const cxxtype * = nullptr) : DataType() {} \
+                                                              \
+    static const bool is_fixed_type = true;                   \
   }
 
 #endif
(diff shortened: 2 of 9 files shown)


lindsayad:
Well that's embarrassing. For my unit test I selected a class that already has an allgather overload `std::basic_string`

lindsayad:
So I only implemented this for allgather kind of as an initial proof-of-concept, and motivated by the fact that when I do this `GeomReal` stuff I don't want to have to remove items in [`parallel_point_test`](https://github.com/libMesh/libmesh/blob/master/tests/parallel/parallel_point_test.C#L43). I think it should be easily extendable to most `Communicator` methods.

On a somewhat related note. I've already started doing some work on enabling parallel communication of `MetaPhysicL` classes; you can see https://github.com/lindsayad/libmesh/commit/fcdd2f701c6f683202a5fa0340a727c70c6c28f4. It seems most appropriate to me to actually do these specializations in `MetaPhysicL` so that users other than libMesh could benefit from them...but then we'd be introducing a submodule dependency in `MetaPhysicL`. So I'm curious what your thoughts on that are

roystgnr:
> It seems most appropriate to me to actually do these specializations in MetaPhysicL so that users other than libMesh could benefit from them...but then we'd be introducing a submodule dependency in MetaPhysicL. So I'm curious what your thoughts on that are

Definitely don't want TIMPI to be a submodule of MetaPhysicL.  But couldn't it just be an optional configure-time dependency?

lindsayad:
> Definitely don't want TIMPI to be a submodule of MetaPhysicL. But couldn't it just be an optional configure-time dependency?

I like that idea.

roystgnr:
Nah, go right ahead.  IIRC you might want to look at the VexCL dependency as a style guide? ... though definitely speak up if anything about *that* looks dubious.

Is this ready to merge?  If so then let's hit the button so I can rebase #6.

lindsayad:
Yes this is ready!

src/parallel/include/timpi/standard_type.h
```diff
@@ -56,12 +56,11 @@ struct standardtype_dependent_false : std::false_type
  * More complicated data types may need to provide a pointer-to-T so
  * that we can use MPI_Address without constructing a new T.
  */
-template <typename T>
+template <typename T, typename Enable = void>
 class StandardType : public DataType
 {
-  // Get a slightly better compiler diagnostic
-  static_assert(standardtype_dependent_false<T>::value,
-                "Only specializations of StandardType may be used, did you forget to include a header file (e.g. parallel_algebra.h)?");
```
roystgnr: I've got to wonder if there isn't a better way to do this.  Right now TIMPI isn't compatible with libMesh, because the StandardType specializations there all have is_fixed_type undefined, plus we've lost this level of diagnostic and so people who accidentally try to invoke a StandardType algorithm with an unsupported type or a missing header are going to get a much more confusing warning message further down the road.

Should we turn is_fixed_type from a static member variable into a separate templated class?  Then use SFINAE (or the sizeof(incomplete type) trick; I think there's others) to make sure it works on any StandardType other than our default here?

roystgnr: I tried out a couple alternatives but they seem convoluted and fragile.  We'll leave this as is unless someone comes up with better C++-fu than mine.

src/parallel/include/timpi/parallel_implementation.h
```diff
@@ -2734,6 +2735,26 @@ inline void Communicator::allgather(const T & sendval,
...
+{
+  TIMPI_LOG_SCOPE ("allgather()","Parallel");
+
+  timpi_assert(this->size());
+  recv.resize(this->size());
+
+  unsigned int comm_size = this->size();
+  if (comm_size > 1)
+    {
+      std::vector<T> range = {sendval};
```
roystgnr, line 2751: @lindsayad - what was the advantage of making a copy here rather than just using sendval.begin() through sendval.end()?

lindsayad: It's been a while since I looked at this...but I believe `T` is an arbitrary type and thus likely doesn't have `begin` and `end` methods...?

roystgnr: Oh, wait, T here isn't a container, huh, it's an individual object.  We don't actually dispatch to a case that handles a container of dynamic size objects yet.  In the individual object case the copy here is still easy to avoid but it's so trivial we don't care.  :-)  Thanks!

lindsayad: Wait @roystgnr enlighten me in my denseness...how could we avoid the copy here? When I wrote this, it really grated me to make the copy, but I didn't see a way around it. I think I thought about doing:
```
std::vector<const T *> range = {&sendval};
```
but then discarded that idea perhaps because (my memory is hazy) the `allgather_packed_range` implementation as it is now didn't seem to make sense with passing a container of `T *` instead of `T`.

lindsayad: Although that doesn't make sense because we do containers of `Elem *` all the time... :thinking:

roystgnr: The catch is that for those containers, T *is* ```Elem *``` - which is why I concluded "so trivial we don't care" - generally when someone has something needing packing, the heavyweight stuff is all behind a pointer.

But I guess that only applies when T really is a raw pointer or when T is doing some kind of copy-on-write implementation, though, huh?  And even std::string isn't allowed to do CoW these days.

But anyway - to avoid the copy we should be able to just do ```T * begin = &sendval; T * end = begin; ++end;```  Pointers work as ranges, even if they're not fancy enough for kids these days.

roystgnr: If you want to try that out, I'm working on some other TIMPI stuff, so we're going to want a libMesh TIMPI update in the near future regardless.

lindsayad:
Ok, I'm going to create an issue so that I remind myself that this is
there. There's a decent chance I won't get to it before you do a TIMPI
update, and that's ok with me.

PR #9, opened 2020-02-24
Add Communicator::sum(std::unordered_map)

jwpeterson:
The `map` and `unordered_map` `sum()` implementations are basically identical, so they both simply use the private `Communicator::map_sum()` implementation function.

### DIFF
--- a/src/parallel/include/timpi/communicator.h
+++ b/src/parallel/include/timpi/communicator.h
@@ -230,6 +230,13 @@ private:
   // Keep track of duplicate/split operations so we know when to free
   bool _I_duped_it;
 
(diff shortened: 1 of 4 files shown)


src/parallel/include/timpi/communicator.h
```diff
@@ -39,6 +39,7 @@
 // These could be forward declarations if only that wasn't illegal
```
roystgnr: The longer this list gets, the more uncomfortable it makes me to drag a ton of std:: headers into any code that wants to use a Communicator, but I still don't see any better way to do it.

PR #10, opened 2020-02-24
Communicator::broadcast(map): Save one communication

jwpeterson:
All other things being equal, it's probably more efficient to broadcast one slightly-larger-than-necessary (due to padding) message than two smaller messages.

Also adds a unit test of Communicator::broadcast(map)

### DIFF
--- a/src/parallel/include/timpi/parallel_implementation.h
+++ b/src/parallel/include/timpi/parallel_implementation.h
@@ -1766,8 +1766,7 @@ inline void Communicator::broadcast(std::map<T1,T2,C,A> & data,
   if (this->rank() != root_id)
     {
       data.clear();
-      for (const auto & pr : comm_data)
-        data[pr.first] = pr.second;
+      data.insert(comm_data.begin(), comm_data.end());
     }
 }
 
--- a/src/parallel/include/timpi/parallel_implementation.h
+++ b/src/parallel/include/timpi/parallel_implementation.h
@@ -1754,33 +1754,20 @@ inline void Communicator::broadcast(std::map<T1,T2,C,A> & data,
   std::size_t data_size=data.size();
   this->broadcast(data_size, root_id);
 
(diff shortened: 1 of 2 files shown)


src/parallel/include/timpi/parallel_implementation.h
```diff
@@ -1754,33 +1754,20 @@ inline void Communicator::broadcast(std::map<T1,T2,C,A> & data,
...
-  timpi_assert(pair_first.size() == pair_first.size());
+  this->broadcast(comm_data, root_id);
 
   if (this->rank() != root_id)
     {
       data.clear();
-      for (std::size_t i=0; i<pair_first.size(); ++i)
-        data[pair_first[i]] = pair_second[i];
+      for (const auto & pr : comm_data)
+        data[pr.first] = pr.second;
```
roystgnr: Could this for loop just be replaced with data.insert(comm_data.begin(), comm_data.end())?

And if so, did you just leave that in here for me to nitpick so I wouldn't be overly embarrassed by how much cleaner you've made the *rest* of the function?

jwpeterson: Let me see if that works, I don't see any reason why it wouldn't...

jwpeterson: Yep, that worked.

lindsayad:
Has anyone tried to compile this with libMesh yet? In `CheckpointIO::read_header` we broadcast a `std::map<subdomain_id_type, std::string>`. With these updated changes we now call `broadcast(std::pair<subdomain_id_type, std::string>)` as opposed to `broadcast(subdomain_id_type)` and `broadcast(std::string)` separately. We have a specialization for `broadcast(std::string)` but not `broadcast(std::pair<subdomain_id_type, std::string>)`. This leads to instantiating `StandardType<std::string>` which no worky

PR #12, opened 2020-02-24
Parallel testing

roystgnr:
This PR also disables oversized pull unit tests, because apparently that's never worked perfectly but our test coverage of it has been even worse.


jwpeterson:
So you're saying the _parallel_ communication library now has _parallel_ testing... this could be huge if true.

PR #13, opened 2020-02-26
Update configuration to really actually find MPI

lindsayad:


roystgnr:
You've deleted the PETSc MPI detection completely?  That'll make this unmergeable into libMesh.  IIRC we have users, if not CI configurations, that are completely relying on autodetection via PETSc.

lindsayad:
Ok, contrary to my statement over at https://github.com/libMesh/libmesh/pull/2382, we are at least partially successful at snooping MPI through PETSc.

Here's the variables in the post-configure TIMPI Makefile:
```
PETSCINCLUDEDIRS = -I/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/include -I/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt//include -I/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/include -I/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/include
PETSCLINKLIBS = -Wl,-rpath,/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/lib -L/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/lib -Wl,-rpath,/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/lib -L/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/lib -Wl,-rpath,/opt/moose/mpich-3.3/gcc-9.2.0/lib -L/opt/moose/mpich-3.3/gcc-9.2.0/lib -Wl,-rpath,/opt/moose/gcc-9.2.0/lib/gcc/x86_64-pc-linux-gnu/9.2.0 -L/opt/moose/gcc-9.2.0/lib/gcc/x86_64-pc-linux-gnu/9.2.0 -Wl,-rpath,/opt/moose/gcc-9.2.0/lib64 -L/opt/moose/gcc-9.2.0/lib64 -Wl,-rpath,/lib/x86_64-linux-gnu -L/lib/x86_64-linux-gnu -Wl,-rpath,/usr/lib/x86_64-linux-gnu -L/usr/lib/x86_64-linux-gnu -Wl,-rpath,/opt/moose/gcc-9.2.0/lib -L/opt/moose/gcc-9.2.0/lib -lpetsc -lHYPRE -lcmumps -ldmumps -lsmumps -lzmumps -lmumps_common -lpord -lscalapack -lsuperlu_dist -lflapack -lfblas -lparmetis -lmetis -lptesmumps -lptscotchparmetis -lptscotch -lptscotcherr -lesmumps -lscotch -lscotcherr -lm -lX11 -lpthread -lstdc++ -ldl -lmpifort -lmpi -lgfortran -lm -lgfortran -lm -lgcc_s -lquadmath -lrt -lm -lpthread -lz -lstdc++ -ldl
PETSC_ARCH = 
PETSC_CC_INCLUDES = -I/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/include
```
So we get the MPI lib location right. But we don't pick up the header include path. I can look at the installed `libmesh-config` to confirm this:
```
	"--include")
	    # handle legacy include paths when needed.
	    if (test "x$legacyinclude" = "xyes"); then
		return_val="-I${includedir}/libmesh $return_val"
	    fi
	    return_val="-I${includedir}
 	                -pthread -I/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/include -I/opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt//include
                        $return_val"
```
All we have are PETSc include directories, not the MPICH include directories. In practice this seems to never matter because the user is compiling with `CXX=mpicxx` which automatically adds the correct include path for MPI.

>  IIRC we have users, if not CI configurations, that are completely relying on autodetection via PETSc.

How is this? PETSc was compiled with MPI detectable, but when libMesh is compiled MPI is not detectable? It seems weird to me that `ACX_MPI` (done right) wouldn't be able to pick-up MPI just as well as PETSc is able to pick-up MPI

lindsayad:
If we really do have situations where the only way we can detect MPI is with PETSc, then this should be closed. It just seems sad that we introduce all these unnecessary library dependencies into TIMPI. This PR just creates such a beautiful config :smile: If I don't have any MPI environment variable set or even CXX, just an `mpicxx` in my PATH, I get this config summary:

```
-------------------------------------------------------------------------------
Optional Packages for Testing:
  MPI......................... : yes
```
No `timpi_optional_INCLUDES` or `timpi_optional_LIBS` required since they are provided automatically by `mpicxx`. And the libtool dependencies still show up in `libtimpi*.la`:
```
' /opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpicxx.la /opt/moose/gcc-9.2.0/lib/../lib64/libstdc++.la /opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpi.la -lpthread -lrt /opt/moose/gcc-9.2.0/lib/gcc/x86_64-pc-linux-gnu/9.2.0/../../../../lib64/libstdc++.la'
```

roystgnr:
> If we really do have situations where the only way we can detect MPI is with PETSc, then this should be closed.

Will do.  Although I can't find the original conversation, so I may be misremembering the details.  In fact, IIRC the problem wasn't failing to detect MPI, it was detecting the *wrong* MPI.  PETSc had been configured to use /opt/modules/some_mpi/somewhere, and so libMesh used that, but then TIMPI wasn't looking at PETSC and it detected /usr/bin/mpicxx instead, creating conflicts.

IMHO the *right* thing to do would be to pass the PETSc-detected MPI from libMesh configure to TIMPI configure arguments, I just wasn't able to figure out how to do that.

> It just seems sad that we introduce all these unnecessary library dependencies into TIMPI.

It's not really a dependency, it's an interaction, but that said I strongly agree!

lindsayad:
So there is a fundamental problem with the current TIMPI configure, and that's if we want to configure another project for which we want MPI but we don't want PETSc...so like for what I'm trying to do right now in MetaPhysicL. If PETSc is a dependency of TIMPI, then I have to ensure that anything that  links to TIMPI also has to link to PETSc. And if I'm doing something like configuring MetaPhysicL with a `future-timpi-dir` then I have no way of configuring `MetaPhysicL` with PETSc unless I add a `petsc.m4` to MetaPhysicL!

lindsayad:
> Will do. Although I can't find the original conversation, so I may be misremembering the details. In fact, IIRC the problem wasn't failing to detect MPI, it was detecting the _wrong_ MPI. PETSc had been configured to use /opt/modules/some_mpi/somewhere, and so libMesh used that, but then TIMPI wasn't looking at PETSC and it detected /usr/bin/mpicxx instead, creating conflicts.

You aren't referring to [this debacle](https://github.com/idaholab/moose/pull/14464#issuecomment-563523582) are you?

lindsayad:
With current master TIMPI:
```
(base) lindad@pop-os:~/projects/moose/libmesh/build/contrib/metaphysicl/test/.libs(metaphysicl-update)$ ldd compare_types_unit 
	linux-vdso.so.1 (0x00007fffec7ef000)
	libmetaphysicl.so.6 => /home/lindad/projects/moose/scripts/../libmesh/installed/lib/libmetaphysicl.so.6 (0x00007f8864d75000)
	libtimpi_opt.so.0 => /home/lindad/projects/moose/scripts/../libmesh/installed/lib/libtimpi_opt.so.0 (0x00007f8864d68000)
	libpetsc.so.3.11 => not found
	libsuperlu_dist.so.6 => not found
	libparmetis.so => not found
	libmetis.so => not found
	libX11.so.6 => /lib/x86_64-linux-gnu/libX11.so.6 (0x00007f8864c02000)
	libmpifort.so.12 => /opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpifort.so.12 (0x00007f8864980000)
	libgfortran.so.5 => /opt/moose/gcc-9.2.0/lib64/libgfortran.so.5 (0x00007f88644f0000)
	libquadmath.so.0 => /opt/moose/gcc-9.2.0/lib64/libquadmath.so.0 (0x00007f88642a9000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f886428d000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f8864285000)
	libmpicxx.so.12 => /opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpicxx.so.12 (0x00007f8864063000)
	libstdc++.so.6 => /opt/moose/gcc-9.2.0/lib64/libstdc++.so.6 (0x00007f8863c84000)
	libmpi.so.12 => /opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpi.so.12 (0x00007f88636ff000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f88636dc000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f88636d1000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f8863580000)
	libgcc_s.so.1 => /opt/moose/gcc-9.2.0/lib64/libgcc_s.so.1 (0x00007f8863368000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8863177000)
	libpetsc.so.3.11 => /opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/lib/libpetsc.so.3.11 (0x00007f8861872000)
	libsuperlu_dist.so.6 => /opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/lib/libsuperlu_dist.so.6 (0x00007f88615a7000)
	libparmetis.so => /opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/lib/libparmetis.so (0x00007f8861367000)
	libmetis.so => /opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/lib/libmetis.so (0x00007f88610fd000)
	libxcb.so.1 => /lib/x86_64-linux-gnu/libxcb.so.1 (0x00007f88610d4000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f8864d7c000)
	libgomp.so.1 => /opt/moose/gcc-9.2.0/lib64/libgomp.so.1 (0x00007f8860e9e000)
	libXau.so.6 => /lib/x86_64-linux-gnu/libXau.so.6 (0x00007f8860e96000)
	libXdmcp.so.6 => /lib/x86_64-linux-gnu/libXdmcp.so.6 (0x00007f8860e8e000)
	libbsd.so.0 => /lib/x86_64-linux-gnu/libbsd.so.0 (0x00007f8860e74000)
```
With this PR TIMPI:
```
(base) lindad@pop-os:~/projects/moose/libmesh/build/contrib/metaphysicl/test/.libs(metaphysicl-update)$ ldd compare_types_unit 
	linux-vdso.so.1 (0x00007ffed578e000)
	libmetaphysicl.so.6 => /home/lindad/projects/moose/scripts/../libmesh/installed/lib/libmetaphysicl.so.6 (0x00007f4adb6a0000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f4adb674000)
	libtimpi_opt.so.0 => /home/lindad/projects/moose/scripts/../libmesh/installed/lib/libtimpi_opt.so.0 (0x00007f4adb667000)
	libmpicxx.so.12 => /opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpicxx.so.12 (0x00007f4adb445000)
	libmpi.so.12 => /opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpi.so.12 (0x00007f4adaec0000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f4adae9b000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f4adae90000)
	libstdc++.so.6 => /opt/moose/gcc-9.2.0/lib64/libstdc++.so.6 (0x00007f4adaab1000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f4ada962000)
	libgcc_s.so.1 => /opt/moose/gcc-9.2.0/lib64/libgcc_s.so.1 (0x00007f4ada74a000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f4ada559000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f4adb6a7000)
	libgfortran.so.5 => /opt/moose/gcc-9.2.0/lib64/libgfortran.so.5 (0x00007f4ada0c7000)
	libquadmath.so.0 => /opt/moose/gcc-9.2.0/lib64/libquadmath.so.0 (0x00007f4ad9e80000)
```

roystgnr:
> If PETSc is a dependency of TIMPI,

Ah, I see the equivocation here.  I was using dependency in the "thing you have to have or you can't configure" sense, you in the "thing that *can* become a link-time dependency" sense.

> then I have to ensure that anything that links to TIMPI also has to link to PETSc.

Yeah, that's a problem.  My usual solution is to create a 'foo-config --libs' that gets incorporated at link time via foo.m4, but that's obviously overkill if we don't actually expect to have any *real* dependency libs, and as you say that doesn't work unless we actually have foo-config available (and thus foo configure done) in time for foo's user's configure to use.

> This was due to TIMPI throwing in a libc++ that conflicted with the libc++ we truly wanted (because TIMPI was throwing in 'L/usr/lib on the beginning of the link line)

That must have been it, yeah.

But if we just leave those flags out, does that necessarily fix things?  Aside from breaking anyone who isn't using a compiler wrapper, the wrappers aren't generally doing anything for linking besides adding "-L" (and apparently "-rpath" with newer wrappers) to the underlying compile line themselves.  Was our TIMPI mpi.m4 getting the wrong wrapper somehow?

roystgnr:
Would it be possible to incorporate just the MPI settings from PETSc's config without pulling in everything?  Use their CXX setting to get their mpicxx?  Or steal the flags from their MPICXX_SHOW?

Or would it be reasonable to use your mpi.m4 here in libMesh as well, and then scream and die in configure there if we find that PETSc's MPI doesn't seem to match?  My big objection is just to the two potentially getting out of sync again.

lindsayad:
>  Aside from breaking anyone who isn't using a compiler wrapper, the wrappers aren't generally doing anything for linking besides adding "-L" (and apparently "-rpath" with newer wrappers) to the underlying compile line themselves.

If someone doesn't have `mpicxx` in their PATH right now and they don't specify either `--with-mpi=DIR` or have MPIHOME set, they simply won't get MPI support because neither `PETSC_INCLUDE_DIRS` nor `PETSC_CC_INCLUDES` actually has a path to the MPI headers. PETSc itself relies on the built-in `-I` in the `mpicc/mpicxx` wrappers

lindsayad:
> Was our TIMPI mpi.m4 getting the wrong wrapper somehow?

In that MOOSE debacle, TIMPI was using the same wrapper that PETSc used. It's just that TIMPI totally unnecessarily threw `-L/usr/lib` and `-I/usr/include` into the flags even though those locations had nothing to do with MPI. The existence of those flags was what introduced the erroneous libc++. That's why in this PR I removed any reference to `/usr` (see [here](https://github.com/libMesh/TIMPI/pull/13/files#diff-84f0f4f40c6fe32a7c6b3736ac0b547fL7) and [here](https://github.com/libMesh/TIMPI/pull/13/files#diff-84f0f4f40c6fe32a7c6b3736ac0b547fL18)).

> Would it be possible to incorporate just the MPI settings from PETSc's config without pulling in everything? Use their CXX setting to get their mpicxx? Or steal the flags from their MPICXX_SHOW?

I thing this is a promising idea. We'd have to use MPICXX_SHOW I think because all they put for `CXX` is `CXX = mpicxx`

roystgnr:
>  It's just that TIMPI totally unnecessarily threw -L/usr/lib and -I/usr/include into the flags even though those locations had nothing to do with MPI.

I think I get it.  Our ancient mpi.m4 tries /usr as a fallback for $MPIHOME, and keeps it when it works, but it isn't working because /usr/bin/mpicxx is a thing but because $CXX was already set to a different mpicxx and so testing the combination of $CXX and -L/usr/lib and so forth worked?

I'm leaning more and more toward "would it be reasonable to use your mpi.m4 here in libMesh as well".  We made our mpi.m4 secondary to petsc.m4 in libMesh because our mpi.m4 was inferior, but perhaps that was the wrong solution, when what we should have done was make our mpi.m4 stop being inferior!

> We'd have to use MPICXX_SHOW I think because all they put for CXX is CXX = mpicxx

My first reaction: but that's enough!  If they're configured with /opt/module/fancy/bin/mpicxx and we try to run with a different mpicxx then we can hopefully detect the distinction (or can we?  Is the --showme option standard?) and warn the user.

lindsayad:
Ok, thinking a little more on this...there are two scenarios:

1) TIMPI and libMesh are configured at different times. libMesh gets its MPI from PETSc. So we want to make sure the MPIs are in sync, so we will also make sure that TIMPI gets its MPI from PETSc. But what if the user uses different PETSC_DIRs? Is this more or less likely than the user using something like different `mpicxx` wrappers?

2) TIMPI and libMesh are configured at the same time. As of right now (and this PR doesn't change this) libMesh and TIMPI both set its CXX compiler based on:
```
  AS_IF([test "$enablempi" != no],
        [CXX_TRY_LIST="mpicxx mpiCC mpicc $CXX_TRY_LIST"],
        AC_MSG_RESULT([>>> Disabling MPI per user request <<<]))
```
In libMesh when we detect and enable PETSc we actually don't even try to compile an MPI program to see whether CXX and the the MPI lib coming from PETSc are even compatible, (unless we are sure that the ability to compile `PetscInitialize` translates to the ability to compile `MPI_Init`. I honestly don't know, but I would guess not).

lindsayad:
> I think I get it. Our ancient mpi.m4 tries /usr as a fallback for $MPIHOME, and keeps it when it works, but it isn't working because /usr/bin/mpicxx is a thing but because $CXX was already set to a different mpicxx and so testing the combination of $CXX and -L/usr/lib and so forth worked?

The ancient `mpi.m4` would keep those flags as long as an MPI program would compile. I don't have a `/usr/bin/mpicxx`, but since I have an `mpicxx` wrapper (which lives in /opt/moose/mpich/.../bin) which appends -I and -L flags (pointing to /opt/moose/mpich/.../include and /opt/moose/mpich/.../lib) I was able to compile an MPI program successfully. Since that compilation was successful, e.g. `enablempi=yes` the old `mpi.m4` script would throw in the `MPI_INCLUDES_PATH=-I/usr/include MPI_LIBS_PATH=-L/usr/lib` flags even though they were totally unrelated (and this is irrespective of the existence of /usr/bin/mpicxx)

lindsayad:
> My first reaction: but that's enough! If they're configured with /opt/module/fancy/bin/mpicxx and we try to run with a different mpicxx then we can hopefully detect the distinction (or can we? Is the --showme option standard?) and warn the user.

I believe that `-show` is standard, or at least expected to exist by most people. @BarrySmith said [back in 2011](https://lists.mcs.anl.gov/pipermail/petsc-users/2011-February/007907.html):

>   Shouldn't the mpcc and mpfort manage providing the include directories
> and libraries automatically (like everyone elses mpicc etc does?) Seems very
> cumbersome that users need to know they strange directories and include them
> themselves? A real step backwards in usability?

Seems like we could potentially verify (in libMesh) that if we are enabling MPI, then our `$CXX -show` (where $CXX is set in `compiler.m4`) matches PETSc's `$CXX -show`. Moreover if we were ever configuring a libMesh with an already installed TIMPI we could also do the verification of equivalence of `$CXX -show`.

lindsayad:
> matches PETSc's `$CXX -show`

Sorry it would have to be trying to match against PETSc `MPICXX_SHOW` since their `$CXX` is just a program name

roystgnr:
Wait a minute - even if it's not standard, it doesn't *have* to be standard - if it doesn't work on a particular compiler, we'll still get the same error messages from both, so we won't get a false failure.  And if we get a false success, not getting a clear error message at configure time for a problem that will show up at compile time is just an annoyance, not a show-stopper.

So does this sound good to you?  We commit an improved mpi.m4 to libMesh first, then if it works there we sync up and use the same mpi.m4 here, and we try to add some consistency checks to libMesh petsc.m4?

lindsayad:
Cases tested for newest push:

- No `--with-mpi`, no MPI env variables, PETSc configured with `--with-mpi=1,--with-mpi-dir=DIR,--with-mpi-compilers=0`
```
  MPI......................... : yes
  MPI_IMPL.................... : petsc
  timpi_optional_INCLUDES..... : -I/opt/moose/mpich-3.3/clang-9.0.0/include 
  timpi_optional_LIBS......... : -Wl,-rpath,/opt/moose/mpich-3.3/clang-9.0.0/lib -L/opt/moose/mpich-3.3/clang-9.0.0/lib -lmpich -lpthread 
```
- No `--with-mpi`, no MPI env variables, PETSc configured with `--with-mpi=0`
```
  MPI......................... : no
```
- No `--with-mpi`, no MPI env variables, PETSc configured with `--with-mpi=1,--with-mpi-dir=DIR,--with-mpi-compilers=1`
```
Optional Packages for Testing:
  MPI......................... : yes
  MPI_IMPL.................... : petsc
  timpi_optional_INCLUDES..... : -I/opt/moose/mpich-3.3/clang-9.0.0/include 
```
- No `--with-mpi`, no MPI env variables, PETSc configured with `--with-mpi=1,--with-mpi-compilers=1`, mpicxx available in PATH
```
  MPI......................... : yes
  MPI_IMPL.................... : petsc
```
- `MPI_HOME` set with $MPI_HOME/bin containing wrappers
```
  MPI......................... : yes
  MPI_IMPL.................... : mpi-wrapper-built-in
```
- `MPI_HOME` set with no $MPI_HOME/bin
```
  MPI......................... : yes
  MPI_IMPL.................... : mpi
  timpi_optional_INCLUDES..... : -I/opt/moose/mpich-3.3/clang-9.0.0/include 
  timpi_optional_LIBS......... : -L/opt/moose/mpich-3.3/clang-9.0.0/lib -lmpi  
```

roystgnr:
You're still pushing everything to the same branch here?  For some reason my "Reopen" button is greyed out.

roystgnr:
Mouseover says the branch "was force pushed or recreated", though I don't quite see why that should be a show-stopper.

Anyway, though, we should test out the new m4 files as a libMesh PR first. If it kills anything there then we'll want to fix it before updating the TIMPI submodule anyway.

PR #14, opened 2020-02-27
Add back old map broadcast method

lindsayad:
Use SFINAE based on whether the map types are fixed types in order
to determine whether to call @jwpeterson's single broadcast method
or the old double broadcast method

### DIFF
--- a/src/parallel/include/timpi/communicator.h
+++ b/src/parallel/include/timpi/communicator.h
@@ -239,9 +239,25 @@ private:
 
   /**
    * Private implementation function called by the map-based broadcast()
-   * specializations.
+   * specializations. This is_fixed_type variant saves a communication by broadcasting pairs
    */
-  template <typename Map>
+  template <typename Map,
+            typename std::enable_if<StandardType<typename Map::key_type>::is_fixed_type &&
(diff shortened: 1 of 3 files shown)


jwpeterson:
Thanks for fixing this. We will probably also have the same issue with the `sum(map)` function I just added, since it uses a similar approach of communicating pairs, just with `allgather()` instead of broadcast? I will add some non-fixed-type testing for that one and see what happens.

In general, I'm not sure whether the additional complexity of having to maintain fixed and non-fixed versions of all or most Communicator functions is outweighed by the potential performance improvements. I did not do any profiling to see whether the single-broadcast case really is any faster for maps yet...

jwpeterson:
Yeah I confirmed that `Communicator::sum(map<string,int>)` has the same issue, so I'm going to merge this and then add a similar fix for that...

jwpeterson:
Yeah I confirmed that `Communicator::sum(map<string,int>)` has the same issue, so I'm going to merge this and then add a similar fix for that...

jwpeterson:
Yeah I confirmed that `Communicator::sum(map<string,int>)` has the same issue, so I'm going to merge this and then add a similar fix for that...

PR #16, opened 2020-02-27
Make sure that we can test OPROF and PROF methods

lindsayad:
Without this PR  a user who has oprof or prof in their `METHODS` variable will get a test failure because the oprof tests won't be built but the `run_unit_tests.sh` script will try to run them.

On a related note, with the addition of `run_unit_tests.sh` (which I'm very appreciate for) I'm guessing there's really no way with autotools to still get the pass/fail for each individual test...it will just be a single pass/fail for the `run_unit_tests.sh` script?

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -110,5 +110,71 @@ if BUILD_OPT_MODE
   check_PROGRAMS += dispatch_to_packed_unit-opt
 endif
 
+if BUILD_OPROF_MODE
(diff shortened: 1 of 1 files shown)


roystgnr:
I think there are a couple testing frameworks that autotools has built-in support for, but nothing we'd want to use.  We could make run_unit_tests.sh into a "test runner", though, and leave the actual executables in TESTS.  I've never tried that before but it seems straightforward.

PR #17, opened 2020-03-02
Add Communicator::max(map)

jwpeterson:
Similar to broadcast(map) and sum(map), this is based on a shared private Communicator::map_max() implementation so that it works for both std::map and std::unordered_map.

There are two versions of Communicator::map_max(), one for fixed types and one for non-fixed types. The SFINAE principal is used to control which version gets called.

### DIFF
--- a/src/parallel/include/timpi/communicator.h
+++ b/src/parallel/include/timpi/communicator.h
@@ -278,6 +278,28 @@ private:
                      const unsigned int root_id,
                      const bool identical_sizes) const;
 
+  /**
+   * Private implementation function called by the map-based max()
+   * specializations.  This is_fixed_type variant saves a
+   * communication by broadcasting pairs
+   */
+  template <typename Map,
+            typename std::enable_if<StandardType<typename Map::key_type>::is_fixed_type &&
+                                    StandardType<typename Map::mapped_type>::is_fixed_type,
+                                    int>::type = 0>
+  void map_max(Map & data) const;
+
+  /**
+   * Private implementation function called by the map-based max()
+   * specializations.  This !is_fixed_type variant calls allgather
+   * twice: once for the keys and once for the values.
+   */
+  template <typename Map,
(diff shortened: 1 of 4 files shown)


src/parallel/include/timpi/communicator.h
```diff
@@ -276,6 +276,28 @@ class Communicator
...
+   * Private implementation function called by the map-based max()
+   * specializations.  This !is_fixed_type variant calls allgather
+   * twice: once for the keys and once for the values.
+   */
+  template <typename Map,
+            typename std::enable_if<!(StandardType<typename Map::key_type>::is_fixed_type &&
+                                      StandardType<typename Map::mapped_type>::is_fixed_type),
+                                    int>::type = 0>
+  void map_max(Map & data) const;
+
```
roystgnr: Despite how much C++ template and SFINAE syntax suck, the more I see this idiom the more I like it.

jwpeterson:
> But even an allgather based algorithm is a big improvement over no algorithm at all.

OK, yeah I did not put much thought into trying to optimize beyond the fixed/non-fixed key/value cases, so almost certainly there's more efficient ways of achieving the same result.

roystgnr:
Did a rebase on master and a force push, so we would get the fix in Travis CI.  I'll merge this once it passes while getting tested properly.

roystgnr:
Or I'll see failed tests.  I can replicate these, too.

jwpeterson:
> Or I'll see failed tests. I can replicate these, too.

Should be fixed now, the tests only worked in serial previously but I didn't check them carefully enough because I thought they were running in parallel.

roystgnr:
> because I thought they were running in parallel.

You thought the tests for the parallel communications library were being run in parallel, merely because they were being configured to run in parallel and launched from a test harness that had been parallelized in one PR and had parallel CI supposedly turned on in another PR?  Not paranoid enough!

PR #19, opened 2020-03-02
export TIMPI_RUN in Travis MATRIX_EVAL

roystgnr:
Hopefully this way it'll get passed down to our run_unit_tests.sh and
we'll *actually* be running them in parallel.

Simply setting the variable without exporting it in #12 didn't work.

With luck we'll now see the tests fail, as there's an overzealous unit test (or IMHO an insufficiently flexible API) among the ones @jwpeterson introduced in #10.  I'll commit a fix for the API in a separate PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
           packages:
             - g++-7 openmpi-bin openmpi-common libopenmpi-dev
       env:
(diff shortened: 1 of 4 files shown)


roystgnr:
Rebased on #20 - we'll see if Travis is now happy as expected.

PR #20, opened 2020-03-02
Add identical_sizes argument to broadcast

roystgnr:
This lets us optimize away a little communication in cases where
containers like set/map/unordered_map can already be sized for their
expected contents, but more importantly it gives us an excuse to *add*
such communication in the broadcast(vector) case, making our default
behavior there more flexible and more consistent with the other
broadcast and other vector APIs.


roystgnr:
I tested this manually in parallel, and after this gets merged we can rebase #19 and make sure Travis agrees.

roystgnr:
Travis liked it in #19, and it's passing when I use it in libMesh tests.

PR #21, opened 2020-03-03
MPI detection round 2

lindsayad:
Detection order, per @roystgnr's white board:

1. `--with-mpi`
2. `$MPIHOME`/`$MPI_HOME`
3. Snoop from PETSc config
4. `mpicxx` wrappers available in PATH
5. There is no 5


lindsayad:
This is contingent on tests passing in libmesh/libmesh#2456

lindsayad:
Ok @roystgnr this is synce

roystgnr:
Okay, based on the recent failure:

2 should definitely be:

> 2. ```$MPIHOME/$MPI_HOME``` while letting ```$MPI_INCLUDE + $MPI_LIB``` take precedence over ```$HOME/include + $HOME/lib``` if the more specific environment variables exist

And perhaps 0 should be:

> 0. Test ```$CC/$CXX``` and see if they can build an MPI program without us doing any more futzing.

lindsayad:
I agree with both of those

PR #22, opened 2020-03-04
Nonblocking vec<vec> receives + syncs

### LINKED ISSUE #18: Need a non all to all push_parallel_vector_data
Need to implement a `push_parallel_vector_data()` that doesn't use `alltoall()`, similar to how it is done in `push_parallel_packed_range()`.

roystgnr:
This lets us get rid of the parallel_sync.h overloads with the alltoall step, so now every single sync function there uses the NBX algorithm as adapted by @friedmud.  Fixes #18.


roystgnr:
I wish Github would display commit lists here in topological rather than chronological order...

roystgnr:
Oh, but of course I forgot to check the --disable-mpi builds.

PR #23, opened 2020-03-07
automake fixups

lindsayad:
- Add */timpi_config.h to DISTCLEAN. I believe this actually caused
  me a nasty "bug" at one point where I was compiling TIMPI with
  a different config than what I was compiling a downstream app.
  Moreover, this meets the definition of a configure-generated file
  and hence should be included in `make distclean`. This is
  not currently caught by the default `make distclean` target
- Remove the MAINTAINERCLEANFILES targets. According to the automake
  manual: "maintainer-clean should not delete anything that needs to
  exist in order to run configure and make"
- Add prof and oprof targets for `timpi_version`

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -44,21 +44,4 @@ install-data-hook:
 
 # Additional files to be deleted by 'make distclean'
 DISTCLEANFILES  = _configs.sed
-
-# Files to be deleted by 'make maintainer-clean'
-MAINTAINERCLEANFILES  = aclocal.m4
-MAINTAINERCLEANFILES += aminclude.am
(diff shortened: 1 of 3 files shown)


roystgnr:
Damn, I think I've got a lot of projects that aren't following that maintainer-clean advice... although IMHO if you have to leave any bootstrap junk around then I don't see the point, you might as well just always use distclean.

pbauman:
I think that advice depends on how you're shipping your build system. If you're not shipping `configure` built, i.e. you expect the user to bootstrap, then I think that advice is incorrect.

PR #27, opened 2020-04-03
Create a Packing specialization for std::pair

lindsayad:
This specialization is enabled when the std::pair is not of fixed
size, e.g. when either first or second is not of fixed size.
Size of first and second are computed using sizeof when
one is of fixed size, and using Packing::*size when not of fixed
size. I also introduce a couple of helper structs in order to
"automatically" determine the buffer_type for the Packing<pair>
specialization

### DIFF
--- a/src/parallel/include/timpi/packing.h
+++ b/src/parallel/include/timpi/packing.h
@@ -27,6 +27,8 @@
 #include <cstddef>
 #include <iterator>
 #include <vector>
+#include <utility>
+#include <cstring>
 
 
 // FIXME: This *should* be in TIMPI namespace but we have libMesh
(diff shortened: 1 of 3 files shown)


src/parallel/include/timpi/packing.h
```diff
@@ -76,6 +78,210 @@ class Packing {
   static T unpack(BufferIter in, Context * ctx);
 };
 
+// Idiom taken from https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Member_Detector
```
roystgnr, line 81: Damn, I'd forgotten having seen this before.  We should have used it in StandardType instead of forcing everything to retrofit is_fixed_type=true

roystgnr:
This is simply excellent.  I like that you didn't throw in tuple *immediately*; that would just have been showing off.

PR #28, opened 2020-04-21
Don't swap LIBS and CPPFLAGS

roystgnr:
This may be at the root of how a bad CPPFLAGS setting was getting
incorporated into libMesh builds using Conda

### DIFF
--- a/m4/mpi.m4
+++ b/m4/mpi.m4
@@ -244,8 +244,8 @@ AS_IF(
                   enablempi=no
                 ])
     AC_LANG_POP([C++])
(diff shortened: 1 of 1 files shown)


roystgnr:
Going to merge this without review so we can get an updated version into libMesh too immediately.  But although I'm pretty sure "don't swap LIBS and CPPFLAGS" is the right thing to do, we could probably use a proper review rather than just my quick skimming and grepping to make sure there are no similar bugs elsewhere that came out of the MPI detection refactor.

lindsayad:
Thanks @roystgnr

PR #29, opened 2020-04-24
Common m4

roystgnr:
After thinking about https://github.com/libMesh/libmesh/issues/2459, it occurs to me that we might as well make our common autotools submodule actually *be* an autotools submodule, not just a parallelism submodule because that's our deepest dependency.  Thoughts?


lindsayad:
Basically you're saying any `m4` shared by our repos is a candidate for being put in the new repo, not just stuff related to MPI/parallelism? Yea I mean anything that can eliminate code duplication sounds like a good thing to me (although maybe then we should nest the autoconf archive in libmesh/autoconf-submodule? :rofl: )

PR #30, opened 2020-04-29
Use a relative path for submodules

roystgnr:
Thanks to @jwpeterson for pointing out the issue in
https://github.com/libMesh/libmesh/pull/2534

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "m4/autoconf-submodule"]
 	path = m4/autoconf-submodule
-	url = git@github.com:libMesh/autoconf-submodule
+	url = ../../libMesh/autoconf-submodule


roystgnr:
So... did we do something that screwed up Travis CI posts to TIMPI PRs?  Looking at https://travis-ci.org/github/libMesh/TIMPI I see that the tests ran (and passed) for this PR, but looking at https://github.com/libMesh/TIMPI/pulls?q=is%3Apr+is%3Aclosed it appears that none of the Travis results have been posted to GitHub for at least a month.

roystgnr:
Looks like something to do with https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798/12 perhaps?  I've tried setting up the new GitHub Apps integration, but on https://github.com/organizations/libMesh/settings/installations/8474191?repository_ids= I don't actually see TIMPI...

roystgnr:
Ah.  There's a "migration", and we need to sign this account up for the beta to use it... trying to do that, but in any case the problem's not on our end.

PR #34, opened 2020-05-16
More dispatches to packed range; displace is_fixed_type idiom

lindsayad:
@roystgnr has expressed his dislike of the `is_fixed_type` idiom, and I
never loved it either. So let's use the member detector idioms
that are already well known in C++. Something should be sendable
using MPI_Datatype if the `StandardType<T>` specialization inherits
from `Datatype` which has the `_datatype` member indicating an
MPI_Datatype. This kind of thinking requires that `StandardType`
specializations which are not sendable via `MPI_Datatype` **do not**
inherit from `Datatype`. Removing that inheritance also makes it impossible
to call a `Communicator` API that takes a `TIMPI::Datatype` with an object that actually doesn't
support sending via `MPI_Datatype`.

Moreover, I don't think we should enable templates just because something
**is not** true. We should only emplate templates when something **is**
true. So now `MPI_Datatype` APIs are enabled if `Has_datatype` is `true`.
And packed-range dispatching APIs are enabled if `Has_buffer_type` is
`true`.

I've moved quite a few methods out of the `#ifdef TIMPI_HAVE_MPI`
section of `parallel_implementation.h` if they present easy return
opportunities when `comm.size() == 1`. I actually had to do this
because my decision to get rid of enabling of templates based on something
not being `true` actually made it so that there were no viable template
candidates in many of our test cases. I actually found that encouraging.
We should not be falling into an overload by accident.

### DIFF
--- a/src/algorithms/include/timpi/parallel_sync.h
+++ b/src/algorithms/include/timpi/parallel_sync.h
@@ -162,6 +162,7 @@ void push_parallel_vector_data(const Communicator & comm,
   // This function must be run on all processors at once
   timpi_parallel_only(comm);
 
+#ifdef TIMPI_HAVE_MPI
   // This function implements the "NBX" algorithm from
   // https://htor.inf.ethz.ch/publications/img/hoefler-dsde-protocols.pdf
 
@@ -190,6 +191,7 @@ void push_parallel_vector_data(const Communicator & comm,
 
   // The send requests
   std::list<Request> reqs;
+#endif
 
   processor_id_type num_procs = comm.size();
 
@@ -205,9 +207,13 @@ void push_parallel_vector_data(const Communicator & comm,
         act_on_data(destid, datum);
       else
         {
+#ifdef TIMPI_HAVE_MPI
           Request sendreq;
           comm.send(destid, datum, type, sendreq, tag);
           reqs.push_back(sendreq);
+#else
+          timpi_error_msg("We should not be trying to send data without MPI");
+#endif
         }
     }
 
@@ -215,6 +221,7 @@ void push_parallel_vector_data(const Communicator & comm,
   if (comm.size() == 1)
     return;
 
+#ifdef TIMPI_HAVE_MPI
   bool sends_complete = reqs.empty();
   bool started_barrier = false;
   Request barrier_request;
@@ -313,6 +320,9 @@ void push_parallel_vector_data(const Communicator & comm,
 
   // Reset the send mode
   const_cast<Communicator &>(comm).send_mode(old_send_mode);
+#else
+  timpi_error_msg("We should not be trying to receive data without MPI");
+#endif
 }
 
 
--- a/src/algorithms/include/timpi/parallel_sync.h
+++ b/src/algorithms/include/timpi/parallel_sync.h
@@ -162,7 +162,6 @@ void push_parallel_vector_data(const Communicator & comm,
   // This function must be run on all processors at once
   timpi_parallel_only(comm);
 
-#ifdef TIMPI_HAVE_MPI
   // This function implements the "NBX" algorithm from
   // https://htor.inf.ethz.ch/publications/img/hoefler-dsde-protocols.pdf
 
@@ -191,7 +190,6 @@ void push_parallel_vector_data(const Communicator & comm,
 
   // The send requests
   std::list<Request> reqs;
-#endif
 
   processor_id_type num_procs = comm.size();
 
@@ -207,13 +205,9 @@ void push_parallel_vector_data(const Communicator & comm,
         act_on_data(destid, datum);
       else
         {
-#ifdef TIMPI_HAVE_MPI
           Request sendreq;
           comm.send(destid, datum, type, sendreq, tag);
           reqs.push_back(sendreq);
-#else
-          timpi_error_msg("We should not be trying to send data without MPI");
-#endif
         }
     }
 
@@ -221,7 +215,6 @@ void push_parallel_vector_data(const Communicator & comm,
   if (comm.size() == 1)
     return;
 
-#ifdef TIMPI_HAVE_MPI
   bool sends_complete = reqs.empty();
   bool started_barrier = false;
   Request barrier_request;
@@ -320,9 +313,6 @@ void push_parallel_vector_data(const Communicator & comm,
 
   // Reset the send mode
   const_cast<Communicator &>(comm).send_mode(old_send_mode);
-#else
-  timpi_error_msg("We should not be trying to receive data without MPI");
-#endif
 }
 
 
--- a/src/parallel/include/timpi/communicator.h
+++ b/src/parallel/include/timpi/communicator.h
@@ -21,11 +21,11 @@
 
 // TIMPI includes
 #include "timpi/standard_type.h"
+#include "timpi/packing.h"
 #include "timpi/message_tag.h"
 #include "timpi/timpi_config.h"
 #include "timpi/request.h"
 #include "timpi/status.h"
-#include "timpi/packing_forward.h"
 
 // C++ includes
 #include <map>
@@ -45,6 +45,7 @@ namespace TIMPI
 {
 
 using libMesh::Parallel::Packing;
+using libMesh::Parallel::Has_buffer_type;
 
 // Define processor id storage type.
 #if TIMPI_PROCESSOR_ID_BYTES == 1
@@ -239,8 +240,8 @@ private:
    * by broadcasting pairs
    */
   template <typename Map,
-            typename std::enable_if<StandardType<typename Map::key_type>::is_fixed_type &&
-                                    StandardType<typename Map::mapped_type>::is_fixed_type,
+            typename std::enable_if<Has_datatype<StandardType<typename Map::key_type>>::value &&
+                                    Has_datatype<StandardType<typename Map::mapped_type>>::value,
                                     int>::type = 0>
   void map_sum(Map & data) const;
 
@@ -250,8 +251,8 @@ private:
    * twice: once for the keys and once for the values.
    */
   template <typename Map,
-            typename std::enable_if<!(StandardType<typename Map::key_type>::is_fixed_type &&
-                                      StandardType<typename Map::mapped_type>::is_fixed_type),
+            typename std::enable_if<!(Has_datatype<StandardType<typename Map::key_type>>::value &&
+                                      Has_datatype<StandardType<typename Map::mapped_type>>::value),
                                     int>::type = 0>
   void map_sum(Map & data) const;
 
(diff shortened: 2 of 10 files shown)


roystgnr:
Ping me when the unit tests are in?  Code looks reasonable but I've been having a maddening time with everything electronic this week and right now I don't dare merge something that hasn't been heavily tested.

lindsayad:
Ok @roystgnr over five months later, I've added my unit tests

lindsayad:
I can't decide whether to these older compilers are dumb for not realizing that there's no use of `DataType::_datatype` in `push_parallel_vector_data` or smart for getting at the thing that I feel worst about in this PR: trying to trick-build a `StandardType<dynamic_type>` so that I can use the same code-paths in `push_parallel_vector_data` and `pull_parallel_vector_data`

lindsayad:
These older compilers can't figure out that `Has_datatype` should be `true`...?

lindsayad:
Ok `Has_datatype<StandardType<T>>` replaced by `std::is_base_of<DataType,StandardType<T>>`

lindsayad:
This is ready for review

src/algorithms/include/timpi/parallel_sync.h
```diff
@@ -205,16 +207,21 @@ void push_parallel_vector_data(const Communicator & comm,
         act_on_data(destid, datum);
       else
         {
+#ifdef TIMPI_HAVE_MPI
           Request sendreq;
           comm.send(destid, datum, type, sendreq, tag);
           reqs.push_back(sendreq);
+#else
+          timpi_error_msg("We should not be trying to send data without MPI");
```
roystgnr, line 215: We already had ```timpi_not_implemented()``` in the serial_implementation.h ```send()```, right?  That will give a message and throw an exception up the stack at run time, so isn't that sufficient?

src/algorithms/include/timpi/parallel_sync.h
```diff
@@ -162,6 +162,7 @@ void push_parallel_vector_data(const Communicator & comm,
   // This function must be run on all processors at once
   timpi_parallel_only(comm);
 
+#ifdef TIMPI_HAVE_MPI
```
roystgnr, line 165: Was this failing to compile without before, or was it failing to run (in the trivial cases where one rank asks to talk to itself)?  If so how was it passing CI?  If not, why bother ifdef'ing it out?  I certainly want --disable-mpi to remain *supported* indefinitely, but I'm hoping to minimize how much cruft it adds to the code.

src/algorithms/include/timpi/parallel_sync.h
lindsayad: Yea I suppose. I guess I tend to be paranoid and assume that someone evil might come along and remove errors in methods down the stack. But then again perhaps that evil person is not evil at all and has actually implemented something sensical to do when someone asks to send data in serial...

src/parallel/include/timpi/communicator.h
```diff
@@ -239,8 +240,8 @@ class Communicator
    * by broadcasting pairs
    */
   template <typename Map,
-            typename std::enable_if<StandardType<typename Map::key_type>::is_fixed_type &&
-                                    StandardType<typename Map::mapped_type>::is_fixed_type,
+            typename std::enable_if<std::is_base_of<DataType, StandardType<typename Map::key_type>>::value &&
```
roystgnr, line 243: This was really smart.  Although I like to think that I might have thought of it too, if I'd remembered that is_base_of existed.

src/parallel/include/timpi/communicator.h
```diff
@@ -589,14 +608,35 @@ class Communicator
...
                          const MessageTag & tag) const;
 
+  /**
+   * Nonblocking-receive from one processor with user-defined type. Dispatches to \p
+   * possibly_receive_packed_range
+   *
+   * @param src_processor_id The pid to receive from or "any".
+   * will be set to the actual src being received from
+   * @param buf The buffer to receive into
+   * @param type The intrinsic datatype to receive
```
roystgnr, line 626: This isn't an intrinsic datatype anymore, though, right?

src/parallel/include/timpi/data_type.h
```diff
@@ -104,6 +105,30 @@ class DataType
   data_type _datatype;
 };
 
+class NotADataType
```
roystgnr, line 108: Let's add some docs for this?  It makes sense but not at first glance.

src/parallel/include/timpi/packing.h
```diff
@@ -389,7 +389,7 @@ inline Iter pack_range (const Context * context,
 #endif
 
       Packing<T>::pack
-        (*range_begin, back_inserter(buffer), context);
+        (*range_begin, std::back_inserter(buffer), context);
```
roystgnr, line 392: Damn, nice catch.  How did this compile before?  Because buffer was always some std::container and so ADL applied?

src/parallel/include/timpi/parallel_implementation.h
```diff
@@ -3123,7 +2930,59 @@ inline void Communicator::allgather(std::vector<std::basic_string<T>,A> & r,
 
   if (identical_buffer_sizes)
     {
-      timpi_assert(this->verify(r.size()));
+      if (r.empty())
+        return;
+
+      timpi_assert(this->verify(r.size()));
```
roystgnr, line 2936: Let's do this assert *before* we rely on the assertion being true in that if statement?

src/algorithms/include/timpi/parallel_sync.h
lindsayad: Oh yes... This does actually fail to compile... I believe in `parallel_sync_unit.C` when we `testPushVecVec`. In serial we do not have the `send` specialization for `vector<vector<T>>`. This compiled *before* because we have this definition in `serial_implementation.h`:
```c++
template <typename T>
inline void Communicator::send (const unsigned int,
                                const T &,
                                const DataType &,
                                Request &,
                                const MessageTag &) const
{ timpi_not_implemented(); }
```
But now I've made the default `StandardType` definition no longer inherit from `DataType`, so that `send` overload is no longer a candidate, and in fact there are no candidates. As opposed to adding a new overload for `NotADataType` that would simply implement `timpi_not_implemented` I decided it made more sense to add the `#ifdef TIMPI_HAVE_MPI`. I also think the fact that that particular `send` overload is no longer a candidate is a good thing; I really like that `StandardTypes` that aren't meant to be sent via MPI no longer inherit from `DataType`. That's just my opinion though

src/parallel/include/timpi/packing.h
lindsayad: I committed that back in May and unfortunately I don't even remember how I caught this!

src/parallel/include/timpi/standard_type.h
```diff
@@ -57,11 +57,8 @@ struct standardtype_dependent_false : std::false_type
  * that we can use MPI_Address without constructing a new T.
  */
```
roystgnr, line 58: Let's also document the DataType-vs-NotADataType distinction here?

roystgnr:
It's starting to look really good.  I can't say I trust such deep changes based on Travis CI alone, though.  We really need to get Civet hooked in here.  I assume you've already manually tried it out in libMesh+MOOSE?

lindsayad:
> It's starting to look really good. I can't say I trust such deep changes based on Travis CI alone, though. We really need to get Civet hooked in here. I assume you've already manually tried it out in libMesh+MOOSE?

Testing now. I do think that Travis *should* be enough, in theory. If we catch things by testing libMesh/MOOSE that our TIMPI standalone tests don't catch, then I think that's incentive to increase our test coverage here at the bottom. Although I guess that still relies on doing libMesh/MOOSE testing to find those gaps in the first place :smile:

src/algorithms/include/timpi/parallel_sync.h
lindsayad: Relating to https://github.com/libMesh/TIMPI/pull/34#discussion_r496088334, I think either I should add a `send` overload that has `timpi_not_implemented` and then we can remove the `#ifdef TIMPI_HAVE_MPI` here. Or we keep the `#ifdef` *and* the error message. Because it would be worst case scenario if somehow a user got into that `else` branch and we simply silently did nothing. I'll defer to your judgment on which of those two you would prefer.

jwpeterson:
> We already had timpi_not_implemented()

This is really a missed opportunity to have a `noTIMPLImented()` macro.

lindsayad:
> This is really a missed opportunity to have a `noTIMPLImented()` macro.

Wow, how long have you been holding onto that one?

jwpeterson:
> Wow, how long have you been holding onto that one?

You can't put a timer on inspiration.

src/algorithms/include/timpi/parallel_sync.h
roystgnr: Although in most cases it's a no-brainer that compile-time errors are better than run-time errors, I do think this is an exception.  It's too natural to write code that calls a send() to processors other than one's self, and then that code naturally avoids calling a send() in serial, but it still *has* a call to send() in serial so it would still fail to compile without ifdef wrapping.

If we had more of a 50/50 split of MPI-users vs serial-only-users then I'd say the extra protection for the latter use case was worth the extra hassle for the former, but 95/5 is probably an underestimate.

So let's add the timpi_not_implemented() at the low level, and then any required changes at the high level for correctness in the serial case (e.g. making sure we do a copy rather than a send+receive for to-self data) will probably end up being efficiency improvements in the parallel case too.

roystgnr: I have considered becoming the "not evil at all and has actually implemented something sensical to do when someone asks to send data in serial" person at one point, but it would only make sense for non-blocking sends, so it wouldn't be a complete fix.

lindsayad: By the same token, I should take out the `#idef` wrapping of `possibly_receive`...

lindsayad: I'm kind of thinking that
```c++
template <typename T>
inline void Communicator::send (const unsigned int,
                                const T &,
                                const DataType &,
                                Request &,
                                const MessageTag &) const;
```
actually isn't defined in parallel. Am I missing something?

lindsayad: We must catch all `send` calls with more specific overloads in parallel. So maybe it's intential that we only define this in serial?

roystgnr: Well, huh.  My best guess is that I only created overloads with various containers, because that was all we needed at first, and then when we added the ability to override StandardType<T>() with a user-specified DataType, we never added that to the unspecialized overload?

But yeah, in theory we keep all the "care needs to be taken in parallel but not in serial" specializations wrapped in ifdef TIMPI_HAVE_MPI in parallel_communicator_specializations, and then all the serial implementation needs to define is the non-specialized cases and any cases where the specialization/overload matters in serial too.

roystgnr (approved):

lindsayad:
I'll report back when I'm finished with downstream testing

lindsayad:
Ok I've gotten all passes for:
- MOOSE test debug
- MOOSE modules optimized
- libMesh make check optimized

libMesh make check debug is still running, but everything has passed thus far.

lindsayad:
Ok per https://github.com/libMesh/libmesh/pull/2719 I'm ready to merge this unless someone screams in the next hour or two

PR #33, opened 2020-05-06
push_parallel_packed_range fixes and improvements

roystgnr:
This fixes the possibly-unused-variable warning that had my --enable-werror libMesh build screaming at me, along with some other fixes I saw along the way, a big expansion of compatibility with more container types, and a couple unit tests.

Haven't tested this with the fixed libMesh node constraints code yet, but I'll do that before I hit merge.

PR #36, opened 2020-06-25
Better packed_range buffer size control

roystgnr:
This adds optional approx_buffer_size arguments to as many more packed_range functions as possible, and in the case where it's impossible (allgather dispatch as in #5) tries to choose a buffer size that won't overflow regardless of problem size.

Thanks to @fdkong for tracking this down.  This is the first step of the fix, but we're going to actually have to *use* the new parameter in one or two places in libMesh before his runs will be safe.

### DIFF
--- a/src/parallel/include/timpi/parallel_implementation.h
+++ b/src/parallel/include/timpi/parallel_implementation.h
@@ -586,7 +586,8 @@ inline void Communicator::send_packed_range (const unsigned int dest_processor_i
                                              const Context * context,
                                              Iter range_begin,
                                              const Iter range_end,
-                                             const MessageTag & tag) const
+                                             const MessageTag & tag,
+                                             std::size_t approx_buffer_size) const
 {
   // We will serialize variable size objects from *range_begin to
   // *range_end as a sequence of plain data (e.g. ints) in this buffer
@@ -608,7 +609,7 @@ inline void Communicator::send_packed_range (const unsigned int dest_processor_i
       std::vector<typename Packing<T>::buffer_type> buffer;
 
       const Iter next_range_begin = pack_range
-        (context, range_begin, range_end, buffer);
+        (context, range_begin, range_end, buffer, approx_buffer_size);
 
       timpi_assert_greater (std::distance(range_begin, next_range_begin), 0);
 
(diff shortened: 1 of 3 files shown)


roystgnr:
Forgot the --disable-mpi case.  Yay CI.

roystgnr:
Huh.  Those other --disable-mpi failures were regressions in #33, but for some reason I can see the tests there in Travis logs but we never saw results posted to GitHub.

roystgnr:
This has been rebased on top of the regression fix, and cleaned up in the process; I'll merge tomorrow unless there are any complaints.

src/parallel/include/timpi/parallel_implementation.h
```diff
@@ -3028,12 +3031,17 @@ inline void Communicator::allgather(const T & sendval,
   timpi_assert(this->size());
   recv.resize(this->size());
 
+  static const std::size_t approx_total_buffer_size = 1e8;
```
fdkong, line 3034: I do not know we can write an integer as 1e8. I thought is was a float type.

fdkong:
Looks nice for me. Thanks

src/parallel/include/timpi/parallel_implementation.h
roystgnr: It is, but C++ will silently convert float types to integral types, and if the compiler can prove that the float value fits in the integer exactly (as it can with a compile-time constant) then even -Wconversion won't complain about it.

lindsayad: I assume that there is a motivation for the choice of this number? Just at first glance it seems kind of arbitrary

fdkong: The total message should be less than 1G.

fdkong: when considering today most the supercomputers  will have 2G-4G per core

roystgnr: My thinking exactly, @fdkong.  I basically divided 1GB by a typical sizeof(buffertype) of 8 bytes then shrunk it a little further to be safe.

Though to be sure, I still don't like leaving this out of user control.  If you weren't happy with my still-pretty-arbitrary choice then my next thought would have been to move this from a static local into a Communicator member variable, and I'm still kind of considering it regardless.

PR #37, opened 2020-06-25
Fix regression with --disable-mpi

roystgnr:
This was apparently broken by #33 but our CI wasn't working right.

In the course of testing the actual fix, I also refactored the implementation header slightly to improve searchability, and added "make test_headers" support to make sure I wasn't subtly breaking any headers.


roystgnr:
Merging right away since there's a significant bugfix; we can revisit if anyone doesn't prefer my header refactoring so I can tell them why they're wrong.

PR #38, opened 2020-06-30
Cache StandardType in push_parallel_vector_data

roystgnr:
This avoids destroying and reallocating a StandardType on each NBX loop, which was taking up a ton of time in some of our profiling.  Fingers crossed, this might fix a recent performance regression *and* further improve a recent performance optimization.

On the other hand, with this change we may just end up busy looping more often, rather than actually finishing communications faster.

@fdkong, could you try this out on the same problem you've been profiling for me?  I'd love to see how the gperftools output changes.

### DIFF
--- a/src/algorithms/include/timpi/parallel_sync.h
+++ b/src/algorithms/include/timpi/parallel_sync.h
@@ -169,6 +169,9 @@ void push_parallel_vector_data(const Communicator & comm,
   // without confusing one for the other
   auto tag = comm.get_unique_tag();
 
+  // We'll construct the StandardType once rather than inside a loop
+  auto type = build_standard_type(&data.begin()->second);
+
   // Post all of the sends, non-blocking and synchronous
 
   // Save off the old send_mode so we can restore it after this
@@ -196,7 +199,7 @@ void push_parallel_vector_data(const Communicator & comm,
       else
         {
           Request sendreq;
-          comm.send(destid, datum, sendreq, tag);
+          comm.send(destid, datum, type, sendreq, tag);
           reqs.push_back(sendreq);
         }
     }
@@ -227,7 +230,9 @@ void push_parallel_vector_data(const Communicator & comm,
     current_src_proc = any_source;
 
     // Check if there is a message and start receiving it
-    if (comm.possibly_receive(current_src_proc, *current_incoming_data, *current_request, tag))
+    if (comm.possibly_receive(current_src_proc,
+                              *current_incoming_data, type,
+                              *current_request, tag))
     {
       receive_reqs.emplace_back(current_src_proc, current_request);
       current_request = std::make_shared<Request>();
--- a/src/algorithms/include/timpi/parallel_sync.h
+++ b/src/algorithms/include/timpi/parallel_sync.h
@@ -169,8 +169,11 @@ void push_parallel_vector_data(const Communicator & comm,
   // without confusing one for the other
   auto tag = comm.get_unique_tag();
 
-  // We'll construct the StandardType once rather than inside a loop
-  auto type = build_standard_type(&data.begin()->second);
+  // We'll construct the StandardType once rather than inside a loop.
+  // We can't pass in example data here, because we might have
+  // data.empty() on some ranks, so we'll need StandardType to be able
+  // to construct the user's data type without an example.
+  auto type = build_standard_type(static_cast<nonconst_nonref_type*>(nullptr));
 
   // Post all of the sends, non-blocking and synchronous
 
--- a/src/parallel/include/timpi/standard_type.h
+++ b/src/parallel/include/timpi/standard_type.h
@@ -73,6 +73,49 @@ private:
 };
 
 
+/*
+ * Template metaprogramming to make build_standard_type work nicely
+ * with nested containers
+ */
+template <typename T>
+struct InnermostType
+{
+  typedef T type;
+};
+
+
+template <typename T>
+struct InnermostType<std::vector<T>>
+{
+  typedef typename InnermostType<T>::type type;
+};
+
+
+/*
+ * Returns a StandardType suitable for use with the example data.
+ */
+template <typename T>
+StandardType<T> build_standard_type(const T * example = nullptr)
+{
+  StandardType<T> returnval(example);
+  return returnval;
+}
+
+
+
+/*
+ * Returns a StandardType suitable for use with the data in the
+ * example container.
+ */
+template <typename T>
+StandardType<typename InnermostType<T>::type>
+build_standard_type(const std::vector<T> * example = nullptr)
+{
+  const T * inner_example = (example && !example->empty()) ? &(*example)[0] : nullptr;
+  return build_standard_type(inner_example);
+}
+
+
 
 // ------------------------------------------------------------
 // Declare StandardType specializations for C++ built-in types


fdkong:
> @fdkong, could you try this out on the same problem you've been profiling for me? I'd love to see how the gperftools output changes.

OK, I will try it soon. Thanks!

roystgnr:
@lindsayad - do we not already have an equivalent of that InnermostType metacode anywhere?  I recall doing something similar with the MetaPhysicL containers but you'd think something that simple would have ended up in libMesh proper and would be applicable to std::vector by now.  Also: your opinion on the "build_standard_type" method?  The obvious alternative would be for StandardType<std::vector<T>> to become well-defined, but that would probably wreak havoc with your automatic dispatch tricks.

lindsayad:
You mean this kind of thing?
```
// Idiom taken from https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Member_Detector
template <typename T>
class Has_buffer_type
{
  using Yes = char[2];
  using No = char[1];

  struct Fallback {
    struct buffer_type {};
  };
  struct Derived : T, Fallback {};

  template <typename U> static Yes &test(U *);

  // this template must be more specialized in general than the Yes version because it involves a
  // type-dependent expression...?
  template <typename U> static No &test(typename U::buffer_type *);

public:
  static constexpr bool value = sizeof(test<Derived>(nullptr)) == sizeof(Yes);
};
```
That's in TIMPI `packing.h` but yea we don't have a macro (like the one described in that wiki) setup for generalizing it.

lindsayad:
I don't think that we do. Are you going to be doing more non-container specializations for `InnermostType` anywhere? Glancing quickly through this code, it seems like you could just find-replace `InnermostType<T>` -> `T`

roystgnr:
The only greater good I can think of is that we'd save a few lines in the PR, which isn't nearly worth it.  So unless there's some third solution we haven't thought of, I'll merge this as-is if and when performance logging results justify it.

fdkong:
I can not compile this branch:

```
make[3]: Leaving directory `/home/kongf/workhome/sawtooth/moose/libmesh/build/contrib/metaphysicl'
make[2]: Leaving directory `/home/kongf/workhome/sawtooth/moose/libmesh/build/contrib/metaphysicl'
Making all in timpi
make[2]: Entering directory `/home/kongf/workhome/sawtooth/moose/libmesh/build/contrib/timpi'
make[2]: *** No rule to make target `all'.  Stop.
make[2]: Leaving directory `/home/kongf/workhome/sawtooth/moose/libmesh/build/contrib/timpi'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/home/kongf/workhome/sawtooth/moose/libmesh/build/contrib'
make: *** [all-recursive] Error 1
```

fdkong:
The code seems to have bugs there. My code just crashed, but again it is hard to debug in a large-scale simulation.

fdkong:
Not sure if ` build_standard_type(&data.begin()->second)` makes sense when `data.empty() `is true.

Segfault happens at build_standard_type(&data.begin()->second) when `data.empty() ` is true.

When data is empty, can we still call `data.begin()->second`? `data.begin()` should be something like `00000000000` or some special address that we should not touch???



```
void push_parallel_vector_data(const Communicator & comm,
                               const MapToVectors & data,
                               const ActionFunctor & act_on_data)
{
  // This function must be run on all processors at once
  timpi_parallel_only(comm);

  // This function implements the "NBX" algorithm from
  // https://htor.inf.ethz.ch/publications/img/hoefler-dsde-protocols.pdf

  typedef decltype(data.begin()->second.front()) ref_type;
  typedef typename std::remove_reference<ref_type>::type nonref_type;
  typedef typename std::remove_const<nonref_type>::type nonconst_nonref_type;

  // We'll grab a tag so we can overlap request sends and receives
  // without confusing one for the other
  auto tag = comm.get_unique_tag();

  // We'll construct the StandardType once rather than inside a loop
  auto type = build_standard_type(&data.begin()->second);
```

roystgnr:
I hadn't even thought about the data.empty() case.  Let me see what I can figure out to do there.

roystgnr:
Well, I typically hope to avoid handling that case, because our StandardType stuff sometimes needs *some* datum to pass to MPI calls, and that means either we hand it an example or we make it default-construct an example, and it would be nice to be able to use as many functions as possible with even C++ classes that lack a default constructor ... but in this case I guess there's no way around the problem; any significantly complicated code is occasionally going to have individual ranks that don't want to send anything out during a parallel sync.  So "default-construct" it is.

roystgnr:
Fun fact: on some systems (an early HP-UX or AIX version, I think it was?), ```0x0``` *wasn't* a special address you shouldn't touch - you could actually read from the first page in memory without causing a segfault.  This was great if you wanted your program not to crash right away when you dereferenced NULL, but not so great if you wanted your program to actually work correctly or be easy to debug when it didn't work correctly.

Wait, did I say "fun" fact?  I meant "drove me mad when I was employed to port old Unix codes to Linux" fact.

roystgnr:
Okay, should be ready for you to try again.  Sorry about the bug; thanks for catching it!

fdkong:
> I hadn't even thought about the data.empty() case. Let me see what I can figure out to do there.

We should have at least one unit test in TIMPI to cover this.  data.empty() happens all the time for parallel simulation or even using only one core.

That being said, I would expect TIMPI works properly when people send an empty message around. For example,

std::vector<Real> empty;

communicator.sum(empty);

I should still correctly get an empty back. That means we should not get a segment fault on this.  `sum` is just an example, I would expect all routines can handle this case.

I would really appreciate it if you could double-check TIMPI and add unit tests for this special case.

fdkong:
> any significantly complicated code is occasionally going to have individual ranks that don't want to send anything out during a parallel sync. So "default-construct" it is.

This may happen even the code is simple.

fdkong:
> Fun fact: on some systems (an early HP-UX or AIX version, I think it was?), `0x0` _wasn't_ a special address you shouldn't touch - you could actually read from the first page in memory without causing a segfault. This was great if you wanted your program not to crash right away when you dereferenced NULL, but not so great if you wanted your program to actually work correctly or be easy to debug when it didn't work correctly.
> 
> Wait, did I say "fun" fact? I meant "drove me mad when I was employed to port old Unix codes to Linux" fact.

There is a statement at http://www.cplusplus.com/reference/map/map/begin/


`If the container is empty, the returned iterator value shall not be dereferenced.`

fdkong:
> Okay, should be ready for you to try again. Sorry about the bug; thanks for catching it!

Hahaha, it is fine, but it indeed took a while (4 hours) to get down that point.

fdkong:
I think you may need to think about this one as well:

`typedef decltype(data.begin()->second.front()) ref_type;`

roystgnr:
My porting-from-Unix experience actually predates the C++98 standard.  :-)  Dereferencing NULL was undefined behavior as per the C standard too, even back then, IIRC, but remember that "undefined behavior" doesn't mean "you're guaranteed a nice clean segfault", it means "You're not guaranteed anything; your code could seem to continue working perfectly for ten more minutes and then make [demons fly out of your nose](http://www.catb.org/jargon/html/N/nasal-demons.html)."

That decltype call sure *looks* unsafe (and perhaps it indirectly caused this bug by tricking me into thinking that executing that code would be safe?), but it's actually perfectly safe and legal!  The contents of a decltype argument never get run.  They get evaluated at compile time, and the "return value" is the type of the expression, whatever type the expression value would have had if it *had* been runnable.  If you check out [the cppreference decltype page](https://en.cppreference.com/w/cpp/language/decltype), the very first example expression is dereferencing an undefined pointer!

roystgnr:
Weird - that Travis build passed days ago but the page here just doesn't want to update.  I guess that's better than the build months ago which *failed* but didn't update Github...

fdkong:
It was hard to get 12,288 cores. I, therefore, used 6,144 cores. We can see improvement but not much.

Without this PR:

profile010



With this PR:

profile009

roystgnr:
3% improvement?  We might as well merge this, I guess, but that's really sad to see.  That's probably within measurement error.

roystgnr:
I wish we had a way to see aggregated data from every MPI rank.  It looks like a lot of the time on proc 0 is being spent busy looping waiting for other procs to catch up, but I can't see a way to tell for sure.  If that is the case, though, then we might be out of "easy" fixes here; the remaining ideas I have are all either big API additions (e.g. early exit with a Request or vector thereof to allow more communication+computation overlap) or huge API additions (e.g. a pull_parallel_vector_data specifically designed for loop operation in cases like allgather_recursive_constraints).

Hey, speaking of that - how heavily are you using constraint equations in this test case?  Periodic BCs, Dirichlet BCs, AMR, user constraints, or some mix of the above?  Spending 8 seconds in Communicator::max() suggests either it's having to handle a decent amount of recursion there or I might be doing something wrong in that loop.

fdkong:
> I wish we had a way to see aggregated data from every MPI rank.

We do have data for every rank, and I could generate graph for each rank. But it sounds too much.

fdkong:
> Hey, speaking of that - how heavily are you using constraint equations in this test case? Periodic BCs, Dirichlet BCs, AMR, user constraints, or some mix of the above? Spending 8 seconds in Communicator::max() suggests either it's having to handle a decent amount of recursion there or I might be doing something wrong in that loop.



I am doing this for the moose framework in general.  The biggest issue is that `equation init` time is not scalable. The time increases quadratically from my observation. Fixing the problem size, the time doubles when we double the number of processor cores.

fdkong:
Why the algorithm is designed recursively?  That could imply we have too many communications for certain applications.

roystgnr:
> I am doing this for the moose framework in general.

Well sure, but that particular run isn't ```moose_test-opt -i everything_in_the_framework.i```, it's some specific problem with some specific set of generated constraints.  Knowing how involved the constraints are would give me a better idea of whether this is a bloated code path I should look into or just a hard problem.

> Why the algorithm is designed recursively?

Because in the AMR and periodic BC and user constraint cases, it's possible for DoF x on processor a to be constrained in terms of dof y on processor b, which is itself constrained in terms of dof z on processor c, etc.  We resolve all that at once at setup time with multiple communication steps, so we can avoid using multiple communication steps every single time we want to enforce constraints exactly subsequently.

roystgnr:
> The biggest issue is that equation init time is not scalable. The time increases quadratically from my observation. Fixing the problem size, the time doubles when we double the number of processor cores.

This is definitely a huge problem.  There may be O(N log2 N) terms we can't optimize further, but O(N^2) for any typical problem suggests we're making a mistake somewhere.  Can you pin the problem area(s) down any more precisely in the call graph?

fdkong:
> Because in the AMR and periodic BC and user constraint cases, it's possible for DoF x on processor a to be constrained in terms of dof y on processor b, which is itself constrained in terms of dof z on processor c, etc. We resolve all that at once at setup time with multiple communication steps, so we can avoid using multiple communication steps every single time we want to enforce constraints exactly subsequently.

For the current test case, we have only peridoic BCs. We do not have AMR or Dirichlet BCs

fdkong:
> This is definitely a huge problem. There may be O(N log2 N) terms we can't optimize further, but O(N^2) for any typical problem suggests we're making a mistake somewhere. Can you pin the problem area(s) down any more precisely in the call graph?

Yes, we at least could do a bit

fdkong:
Oh, boundary constraints are not scalable because we have all boundaries are gathered to every single processor?  So libmesh will try to touch every element of all boundaries.

And then more processors won't help. Libmesh check all ghosted elements for constraints?

fdkong:
Let me try my new fix in moose.

roystgnr:
libMesh itself doesn't ghost boundaries to every processor ... but is that the current trick in MOOSE for working around the libMesh requirement that PeriodicBoundary objects be created+assigned before a Mesh is created+distributed?  That *would* be an O(N^2) bottleneck.

Maybe it's time to add the ability to pull ghosted elements by location, with an octree structure or some other spatial partitioning to make it scale?  I'd been hoping MOOSE would eventually use the libMesh PeriodicBoundary idiom and wouldn't need such a pull just for periodic BCs, but there are problems like contact for which it would also be useful, so we'd want to add it eventually, and if it's now your principal scalability bottleneck then sooner would be better than later... but I'd want to make sure that *that's* really the principal bottleneck first.  We need to look deeper than just "init" and verify where the O(N^2) issue really lies.

fdkong:
I will try these changes https://github.com/idaholab/moose/pull/15485 where boundaries are distributed. 

Let us see if it make any diffences

fdkong:
With distributed boundaries:

profile011

fdkong:
BTW, the time in `allgather_recursive_constraints`  was introduced in the past few months. In a half-year-old libmesh, the compute time in `allgather_recursive_constraints` is negligible. It makes me think we introduced some bug into `allgather_recursive_constraints` in the past few months

It is a profile (generated two months or so) from the old libmesh, and you can not do an apple-to-apple comparison, but you can see the compute time in `allgather_recursive_constraints` is tiny

profile001

roystgnr:
33% faster?  Nice!  And clearly the constraint collection is now the remaining bottleneck here.  You're using periodic BCs on all sides of a cube?  That should be require 3 or 4 levels of recursion, so it's not a trivial problem, but it shouldn't be taking 2/3 of your init time.

Can you see how the DofMap::allgather_recursive_constraints time specifically is growing with processor count?  In theory it shouldn't be quadratic, but at this point I don't see how else it could be taking up that much of your time.

fdkong:
> 33% faster? Nice! And clearly the constraint collection is now the remaining bottleneck here. You're using periodic BCs on all sides of a cube? That should be require 3 or 4 levels of recursion, so it's not a trivial problem, but it shouldn't be taking 2/3 of your init time.

Yes, periodic BCs are on all sides of a cube. But I am still puzzled why we did not see that in the old libmesh.

roystgnr:
That's a very good point.  But we haven't actually changed anything inside allgather_recursive_constraints itself...  Is there any possibility that your current libMesh builds have LIBMESH_ENABLE_NODE_CONSTRAINTS turned on in libmesh_config.h but your previous builds did not?

lindsayad:
```
commit a82e8380b2bd7929356ba214a9fe10d4fd65f658
Author: John Peterson <jwpeterson@gmail.com>
Date:   Thu Jun 11 10:20:00 2020 -0500

    Add --enable-nodeconstraint in update_and_rebuild_libmesh.sh
    
    Refs #15389

diff --git a/scripts/update_and_rebuild_libmesh.sh b/scripts/update_and_rebuild_libmesh.sh
index 05d8777569..c75268a0c5 100755
--- a/scripts/update_and_rebuild_libmesh.sh
+++ b/scripts/update_and_rebuild_libmesh.sh
@@ -161,6 +161,7 @@ if [ -z "$go_fast" ]; then
                                    --disable-maintainer-mode \
                                    --enable-petsc-hypre-required \
                                    --enable-metaphysicl-required \
+                                   --enable-nodeconstraint \
                                    $DISABLE_TIMESTAMPS $VTK_OPTIONS $* | tee -a "$SCRIPT_DIR/$DIAGNOSTIC_LOG" || exit 1
 else
   # The build directory must already exist: you can't do --fast for

```
Old @jwpeterson sneaking in a fast one

jwpeterson:
Enable node constraints, they said. It'll improve test coverage they said. :man_shrugging:

fdkong:
What `--enable-nodeconstraint` does? I mean in which cases I should use that?

I will try to rebuild libmesh with removing ` --enable-nodeconstraint`

roystgnr:
Node constraints allow you to do mesh movement on an AMR mesh without worrying about hanging node positions becoming inconsistent with their neighboring nodes, because libMesh would update any constrained Node positions itself.

And originally the code syncing up those constraints was a thinly updated copy of the code syncing up DoF constraints, IIRC ... but it looks like at some point the DoF constraint code might have been better optimized while the Node constraint code wasn't touched?

roystgnr:
Oh, so to answer your question more clearly: if you're moving your mesh, and it's an AMR mesh, and you're not ensuring in user code that hanging nodes are in consistent positions, then you want --enable-nodeconstraints.  If not then you don't need it.

fdkong:
After removing `--enable-nodeconstraints`, we have 

`Initializing equation system .                                                             [  2.78 s]`

For my current test case, I do not use AMR, so it is fine without `nodeconstraints`.

But in general, it is a good idea to optimize `enable-nodeconstraints.`

roystgnr:
Yeah, I'll definitely look into it.  For now, it sounds like we definitely want to merge this - saving a couple seconds wasn't thrilling before, but is a big deal if we've got the total runtime down to a couple seconds.  (yes, I'm unreasonably assuming that the savings weren't mostly in the node constraints code path, leave me to my delusions)

lindsayad:
So it went from ~47 seconds with node constraints to < 3 seconds? Am I reading things right?

fdkong:
> So it went from ~47 seconds with node constraints to < 3 seconds? Am I reading things right?

Correct. 

So I guess @roystgnr  will take a close into this. AMR is something we definitely want to use in large scale simulation in near future.

lindsayad:
Huh I guess I should know how MOOSE is doing adaptivity with displaced mesh without the use of node constraints off the top of my head, but I don't. And I know we have test cases that do adaptivity with a displaced mesh.

lindsayad:
Ok, I do know how MOOSE works with AMR and displaced mesh... It works because of the infamous `enforce_constraints_exactly` class in `PetscNonlinearSolver`. We constrain the hanging node displacement degrees of freedom before we ever displace the mesh and/or compute our residual/Jacobian

lindsayad:
Old @friedmud back in libmesh/libmesh@b1c6a47a0d6b81

fdkong:
OK, then we do not have to use `--enable-nodeconstraints` somehow.

Which tests will be affected by `--enable-nodeconstraints` right now? We may be able to fix them somewhat

roystgnr:
Correct - if you get the node positions consistent yourself (e.g. by setting them from a set of variables that are made consistent via dof constraints) then you don't need libMesh support for node constraints.

So none of the MOOSE tests' accuracy should be affected by --enable-nodeconstraints.  As for which tests' performance is affected ... on the one hand, "all of them", because we don't short circuit out of that sync step unless no ranks have any of any kind of constraint ... but on the other hand, possibly "none of them" in a practical sense, because the problem here is parallel scalability and we only ever run the MOOSE test suite on a relative handful of processors, not on thousands of cores.

lindsayad:
> So none of the MOOSE tests' accuracy should be affected by --enable-nodeconstraints.

Hence why idaholab/moose#15447 went in without a hitch.

> So none of the MOOSE tests' accuracy should be affected by --enable-nodeconstraints. As for which tests' performance is affected ... on the one hand, "all of them", because we don't short circuit out of that sync step unless no ranks have any of any kind of constraint ... but on the other hand, possibly "none of them" in a practical sense, because the problem here is parallel scalability and we only ever run the MOOSE test suite on a relative handful of processors, not on thousands of cores.

:+1:

lindsayad:
Sigh. I just realized that with `build_standard_type` inside of `push_parallel_vector_data`, we cannot make `Point` in libMesh use a possibly dynamic type for its underlying `_coords` and still use `push_parallel_vector_data`.

> Also: your opinion on the "build_standard_type" method? The obvious alternative would be for StandardType<std::vector> to become well-defined, but that would probably wreak havoc with your automatic dispatch tricks.

Maybe I do want `StandardType<dynamic_thing>` to be a well defined thing now...and still somehow do automatic dispatch to the right underlying `Communicator` method somehow.

This PR did end up being a big win right? I lost track after it was realized that `nodeconstraints` affected performance so much.

lindsayad:
>  The obvious alternative would be for StandardType<std::vector> to become well-defined, but that would probably wreak havoc with your automatic dispatch tricks.

Is there any obvious disaster to defining `StandardType<dynamic_thing>`, and setting `is_fixed_type = false`? I'm going to try it...

lindsayad:
`push_parallel_vector_data` and `push_parallel_packed_range` are so close to being identical! With the right automatic dispatches, is there any fundamental reason these couldn't be merged? Perhaps `Context` could be a default template parameter and `context `can be default `nullptr` ? I foresee that this would require adding default `context` arguments to basically all `Communicator` methods though... :slightly_frowning_face:

fdkong:
> This PR did end up being a big win right? I lost track after it was realized that `nodeconstraints` affected performance so much.

Without this PR, there is an obvious bug. The removement of `--enable-nodeconstraints` is a big win

lindsayad:
I guess the goal of this was was to eliminate the 7.66 seconds out of total 96.65 seconds spent in `StandardType::StandardType` ? That's a pretty good goal, knocking 5-10% out of the simulation.

lindsayad:
> Without this PR, there is an obvious bug. The removement of `--enable-nodeconstraints` is a big win

You mean we wouldn't have discovered `--enable-nodeconstraints` without the discussion on this PR? Or are you referring to some other bug?

roystgnr:
> There's not actually any allocation for standard MPI types like MPI_CHAR right?

Right.

> Only for types for which we have to call MPI_Type_*

Right again, but here's where ease-of-use burns us: it's simple to invoke MPI_Type_create with something as simple as ```pair<int, Real>```, without ever explicitly defining a new StandardType specialization outside of TIMPI code.

I originally avoided duplicate allocations in StandardType by doing the create once, then never doing the resulting free, just reusing the derived type every time we needed it again ... but although OpenMPI and Mpich didn't complain about that, Valgrind rightly called us on the memory leak.  So if we want to avoid the deallocate/reallocate cycle again in the future, we're going to need to start registering allocations in TIMPIInit so we can clean up after ourselves when we're done.

PR #50, opened 2020-10-08
Add nonblocking min, max, and sum

loganharbour:
@roystgnr probably better to have the conversation here.

### DIFF
--- a/src/parallel/include/timpi/communicator.h
+++ b/src/parallel/include/timpi/communicator.h
@@ -351,6 +351,14 @@ public:
   bool semiverify(const T * r) const;
 
   /**
+   * Non-blocking minimum of the local value \p r into \p o
+   * with the request \p req.
+   */
+  template <typename T>
+  inline
+  void min(T & r, T & o, Request & req) const;
+
+  /**
    * Take a local variable and replace it with the minimum of it's values
    * on all processors.  Containers are replaced element-wise.
    */
@@ -377,6 +385,14 @@ public:
   inline
   void minloc(std::vector<T,A1> & r,
               std::vector<unsigned int,A2> & min_id) const;
+
+  /**
+   * Non-blocking maximum of the local value \p r into \p o
+   * with the request \p req.
+   */
+  template <typename T>
+  inline
+  void max(T & r, T & o, Request & req) const;
 
   /**
    * Take a local variable and replace it with the maximum of it's values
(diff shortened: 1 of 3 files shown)


src/parallel/include/timpi/communicator.h
```diff
@@ -350,6 +350,14 @@ class Communicator
   inline
   bool semiverify(const T * r) const;
 
+  /**
+   * Non-blocking minimum of the local value \p r into \p o
+   * with the request \p req.
+   */
+  template <typename T>
+  inline
+  void nonblocking_min(T & r, T & o, Request & req) const;
```
roystgnr, line 359: Let's omit nonblocking_ from the names?  That's implied by the Request argument.

roystgnr:
Oh, you want to get me to put my foot in my mouth in *public*, huh?  Well, the relevant bits from earlier:

Our libmesh_call_mpi() turns into a no-op when MPI isn't available.  But in the case of an in-place reduction operation, a no-op is what you want, and that's still true if you're "non-blocking"!  And for a non-in-place reduction operation, the size==1 optimization is just exactly what you want to do without MPI anyways.

The only case that worries me, I guess, is: what happens if we have MPI, and size==1 so we don't bother using the Request object, but we're in a size-agnostic codepath so we try a test() or wait() on the Request object anyway?

We initialize a Request to MPI_REQUEST_NULL ...

"In this case the operation returns immediately with empty status."

"An empty status is a status which is set to
return tag = MPI_ANY_TAG , source = MPI_ANY_SOURCE , error = MPI_SUCCESS , and is
also internally configured so that calls to MPI_GET_COUNT, MPI_GET_ELEMENTS, and
MPI_GET_ELEMENTS_X return count = 0 and MPI_TEST_CANCELLED returns false ."

So that's pretty much what we'd want to happen in serial, yes?

That Knuth quote about "Caution: I have only proved this code correct, I have not tried running it" is coming to mind right now, but it looks like you've at least got a shot at it working.

In the non-MPI case we're fine for Request::wait(); it becomes a no-op returning a default Status ... and Request::test() is just "return true" so that's good too.

loganharbour:
> Oh, you want to get me to put my foot in my mouth in _public_, huh? 

Exactly 😬 

Should be good now.

roystgnr:
This seems to be passing with every configuration I throw at it too, now.

PR #57, opened 2020-11-11
Ensure methods with SFINAE abling can be used in serial

### LINKED ISSUE #55: Some communicator methods aren't available in serial
Due to the report in libmesh/libmesh#2791, we know that `broadcast(std::string)` is missing in serial due to my changes in #34. Some others may not work:

- [x] `broadcast`
- [x] `allgather`
- [x] `possibly_receive`: I already added the SFINAE signatures in `serial_implementation.h` although they are just `timpi_not_implemented`. That was just copy-pasta from @roystgnr who has the comment: `Non-blocking I/O from self to self?`

lindsayad:
The use of `#ifdef TIMPI_HAVE_MPI` in `communicator.h` goes against the current design, but it's certainly the fewest lines of code to fix #55 I think. If there's a cleaner solution, then I'll change things here. The `allgather` implementation in `serial_implementation` is necessary otherwise we'll get undefined references when trying to link the unit tests (with the new `testAllGatherString` test)

Also refs libmesh/libmesh#2791

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2,7 +2,7 @@
 dnl Process this file with autoconf to produce a configure script.
 
 AC_PREREQ(2.61)
-AC_INIT(timpi, 1.3, roystgnr@ices.utexas.edu)
+AC_INIT(timpi, 1.3.0, roystgnr@ices.utexas.edu)
 AC_CONFIG_MACRO_DIR([m4])
 
 AC_CONFIG_HEADER(src/utilities/include/timpi/timpi_config.h.tmp)
--- a/src/parallel/include/timpi/communicator.h
+++ b/src/parallel/include/timpi/communicator.h
@@ -1208,11 +1208,16 @@ public:
    *
    * Fixed variant
    */
-  template <typename T,
-            typename std::enable_if<std::is_base_of<DataType, StandardType<T>>::value, int>::type = 0>
+  template <typename T
+#ifdef TIMPI_HAVE_MPI
+            ,
+            typename std::enable_if<std::is_base_of<DataType, StandardType<T>>::value, int>::type = 0
+#endif
+            >
   inline void broadcast(T & data, const unsigned int root_id=0,
                         const bool identical_sizes=false) const;
 
+#ifdef TIMPI_HAVE_MPI
   /**
    * Take a possibly dynamically-sized local value and broadcast it to all
    * processors.  Optionally takes the \p root_id processor, which specifies the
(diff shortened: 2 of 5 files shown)


lindsayad:
I could move the `Packing broadcast` declaration into `parallel_communicator_specializations` and its definition up in `parallel_implementation.h` into the `#ifdef TIMPI_HAVE_MPI` section at the sacrifice of losing the organizational grouping with the `DataType` SFINAE overload

lindsayad:
BTW I built all the way down the MOOSE stack with this and all (MOOSE) tests pass

lindsayad:
But that of course was with MPI. I'll build libmesh and perhaps run `make check` with `--disable-mpi`

lindsayad:
Ok libMesh with `--disable-mpi` builds and passes `make check`

roystgnr:
> goes against the current design

I'm not liking that either, but I like it more than everything breaking without MPI.  We can merge this if we can't decide on what the "right" solution is.

But first, my first take on "right": should we be just leaving a simple broadcast(T) in communicator.h, and moving the StandardType-vs-Packed rigmarole to parallel_communicator_specializations?  It's not like any of our other foo(T) declarations are considered promises that the method foo will work with every arbitrary UserDefinedParallelUnwareT that a user can throw at it.  The point of the communicator.h vs parallel_communicator_specializations separation was to leave the "what we'd like to promise" and documentation in one clean relatively short header, while all the messy overloads and details necessary to make that work as much as possible went in the ugly long header.  It seems like the packing dispatch overloads belong in the latter, just like the string and map and set and so forth overloads did.

lindsayad:
> But first, my first take on "right": should we be just leaving a simple broadcast(T) in communicator.h, and moving the StandardType-vs-Packed rigmarole to parallel_communicator_specializations?

Won't this create ambiguous overloads with respect to declarations when we are compiling with a parallel configuration? If `T` has a defined `StandardType<T>` that derives from `DataType`, then both the `communicator.h` generic `T` overload will match as well as the SFINAE `DataType` overload. Equivalently if there is a defined `Packing<T>`, then both the generic `T` overload and SFAINE `Packing` overload will match

roystgnr:
The SFINAE doesn't count as more specific in the usual C++ "more specific overloads take precedence" rules?

Damn, no, it doesn't.  At least not in my tests.  And also while writing my tests I came across a webpage declaring a similar overload attempt to be "A common mistake" in the text and ```/* WRONG */``` in the code snippet, so there's that.

We need to figure out a way around this before every method in communicator.h turns from ```void foo(T)``` to ```ridiculous::and_bloated<cpp,syntax> foo(T); ridiculous::and_redundant_looking<cpp,syntax> foo(T);```  But for now I guess working code is still a big improvement over broken code.

lindsayad:
> Damn, no, it doesn't. At least not in my tests. And also while writing my tests I came across a webpage declaring a similar overload attempt to be "A common mistake" in the text and `/* WRONG */` in the code snippet, so there's that.

It was saying that SFINAE signatures like mine are `WRONG`? That's a shot to the heart :laughing: :cry:

PR #61, opened 2021-02-05
Move Packing<string> into sources

lindsayad:
I actually need the string packing implementation to do a set_union on a
`std::map<T, std::string>`. Moreover, we use string packing in several
tests so it seems of general use


roystgnr:
:+1:

Were you talking to @loganharbour recently?  We were just lamenting that your fancy auto-dispatch wasn't working on a complicated nested structure that he wanted to use, because buried in the vector of tuples of whatever were strings.

lindsayad:
Do we not run into trouble with the tuples?

loganharbour:
std::tuple actually works just fine for me - just not with std::string (before this!).

lindsayad:
That's cool. I'm guessing it's because of all the `std::tuple` stuff @roystgnr has at the bottom of `standard_type.h`. However, I would expect it to fail if any of the types held by `std::tuple` are not themselves `StandardTypes` *and* I guess do not having corresponding `Packing` definitions...? You should try your use case and see whether it works after this PR. I'm still skeptical haha, but would love it if it works!

roystgnr:
Oh, you did your fancy stuff in packing.h for pair but not for tuple, huh?  It looks like most of the hard work is done and would be easy to extend, though.

lindsayad:
I'm assuming a :+1: means good-to-merge?

lindsayad:
Also, when we merge this I'm going to want to bootstrap it and then update the libmesh submodule. Thoughts on versioning and/or just going with `pr_bootstrapped`? If we wanted to update versions, I would think that a patch update would be most appropriate. We're not fixing a bug here, and it doesn't feel nearly big enough to uptick major.

roystgnr:
Good to merge, and agreed that this isn't worth a release in itself.  Maybe after we add Packing<tuple> too.

PR #62, opened 2021-02-08
Packing<tuple> specialization

roystgnr:
This gives us a Packing specialization for any tuple that's not completely composed of fixed types, and removes those cases from the previous StandardType specialization, so variable-size data structures with tuples in them can now participate in automatic packing dispatch.

Hoping @lindsayad can look this over too; he did much of the hard work this is based on.

And, if it's not too late, @loganharbour might want to try this out, since he was the first one to try to communicate a vector of tuples with strings in them under the naive theory that just because we supported vectors and tuples and strings that would mean we must support vectors of tuples of strings too.

### DIFF
--- a/src/parallel/include/timpi/packing.h
+++ b/src/parallel/include/timpi/packing.h
@@ -27,6 +27,7 @@
 // C++ includes
 #include <cstring>     // memcpy
 #include <iterator>
+#include <tuple>
 #include <type_traits> // enable_if, is_same
 #include <utility>     // pair
 #include <vector>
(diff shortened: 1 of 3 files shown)


lindsayad:
I'll look at this after department meeting!

src/parallel/include/timpi/packing.h
```diff
@@ -284,6 +285,263 @@ Packing<std::pair<T1, T2>,
...
+            typename Context,
+            typename std::enable_if<!IsFixed<T3>::value, int>::type = 0>
+  static unsigned int packable_size_comp(const T3 & comp, const Context * ctx)
+  {
+    return Packing<T3>::packable_size(comp, ctx);
+  }
+
+  template <typename Context,
+            std::size_t I>
+  static typename std::enable_if<I == sizeof...(Types), unsigned int>::type
```
lindsayad, line 384: Nice

lindsayad (approved):
This is awesome

src/parallel/include/timpi/packing.h
roystgnr: Wish I could take credit for it.  I did a search for something like "how to iterate over a tuple", and felt like an idiot once I saw the phrase "tail recursion".  I never took a real CS class, but I have learned the basics, and I have used them before (for the heterogeneous structures in MetaPhysicL), and I shouldn't have needed reminding about how.

PR #65, opened 2021-02-11
Packing<list, map, unordered_set, etc>

roystgnr:
This should give us still more flexibility with packed_range operations on nested types.

Hoping for a second set of eyes on this one - it was more tricky than I expected, at least w.r.t. map support thanks to those const Key types.  The bug I hit there was subtle, and though I think I fixed it, I wouldn't be offended if we wanted to just take our time or add more tests rather than sending this downstream as fast as possible.  A little delay might also be worthwhile in order to divide up headers some more; I could see an argument for a "packing_map.h", "packing_tuple.h", etc., rather than forcing every "packing.h" user to include all those containers' headers.  "packing_forward.h" is helpful here but maybe not enough.


roystgnr:
The basic libMesh+MOOSE tests in https://github.com/libMesh/libmesh/pull/2860 were happy with this too.  I'll open some issue tickets to talk about what I think still needs improvement, but this PR is at least a step up in capabilities.

src/parallel/include/timpi/packing.h
```diff
@@ -115,7 +133,9 @@ struct PackingMixedType
   template <typename T3>
   struct IsFixed
   {
-    static const bool value = TIMPI::StandardType<T3>::is_fixed_type;
+    static const bool value =
+      TIMPI::StandardType
+      <typename DefaultValueType<T3>::type>::is_fixed_type;
```
lindsayad, line 138: I think we could get a few more types in here 😄

src/parallel/include/timpi/packing.h
```diff
@@ -152,6 +172,16 @@ struct PackingMixedType
...
 
+  template <typename T1, typename T2,
+            typename OutputIter,
+            typename Context,
+            typename std::enable_if<IsFixed<std::pair<T1, T2>>::value, int>::type = 0>
+  static void pack_comp(const std::pair<T1, T2> & comp, OutputIter data_out, const Context * ctx)
+  {
+    pack_comp(comp.first, data_out, ctx);
+    pack_comp(comp.second, data_out, ctx);
+  }
```
lindsayad, line 183: I'm guessing the general `!IsFixed<T3>` must get called when either `T1` or `T2` here aren't fixed

lindsayad (approved):
Very nice. It makes me feel good to see the same idioms used in the basic `Packing<std::pair>` reused both for tuple and now here.

It looks like you didn't catch the `multis` (`multiset, multimap, unordered_multiset, unordered_multimap`) in testing, but clearly the same idioms tested in the other containers should also work for those.

roystgnr:
> It looks like you didn't catch the multis (multiset, multimap, unordered_multiset, unordered_multimap) in testing

...

No, I didn't.  I think I was distracted by other work and just totally missed it.  I'll add some test coverage there before merging.

src/parallel/include/timpi/packing.h
roystgnr: In theory this case only gets called if ```IsFixed<pair<T1,T2>>```, which only is true for fixed ```T1``` and ```T2```, right.  In practice, this is the sort of thing that scares me a lot.

In some now-rebased-away history, I had done the ```DefaultValueType``` stuff on unpacking (where it was necessary to compile the ```map``` cases, because if ```value_type``` is ```pair<const K, V>``` then you can't write to a ```value_type```) but not on packing (because why bother, you can always *read* a const) ... and this caused the unpacking to try to use fixed-size code (because ```StandardType``` exists for ```pair<K,V>``` iff it exists for ```K``` and ```V```) while the packing tried to use variable-size code with size headers (because ```StandardType<const K>``` doesn't exist).  Some of that extra ```timpi_assert``` foo in the unpacking isn't there because of forethought, it's there because I needed it to figure out what the heck was happening.

roystgnr:
I'm glad I did.  The original ```Avoid memcpy(std::pair)``` commit had a couple bugs in it that were cancelling each other out - one preventing the new unpack_comp specialization from being chosen, and one making it break if it was chosen.  Both are fixed in a fixup commit now.  I'll merge this and make a libMesh update PR once it's through Travis CI again.

PR #66, opened 2021-03-03
Generalized push_parallel_[packed_range/vector_data]

loganharbour:
Ping @roystgnr

### DIFF
--- a/src/algorithms/include/timpi/parallel_sync.h
+++ b/src/algorithms/include/timpi/parallel_sync.h
@@ -184,7 +184,7 @@ push_parallel_nbx_helper(const Communicator & comm,
   const_cast<Communicator &>(comm).send_mode(Communicator::SYNCHRONOUS);
 
   // The send requests
-  std::list<Request> reqs;
+  std::list<Request> requests;
 
   processor_id_type num_procs = comm.size();
 
@@ -200,8 +200,8 @@ push_parallel_nbx_helper(const Communicator & comm,
         act_on_data(destid, datum);
       else
         {
-          reqs.emplace_back();
-          send_functor(comm, destid, datum, reqs.back(), tag);
+          requests.emplace_back();
+          send_functor(comm, destid, datum, requests.back(), tag);
         }
     }
 
@@ -210,7 +210,7 @@ push_parallel_nbx_helper(const Communicator & comm,
     return;
 
   // Whether or not all of the sends are complete
-  bool sends_complete = reqs.empty();
+  bool sends_complete = requests.empty();
 
   // Whether or not the nonblocking barrier has started
   bool started_barrier = false;
@@ -218,7 +218,7 @@ push_parallel_nbx_helper(const Communicator & comm,
   Request barrier_request;
 
   // The pair of src_pid and requests
-  std::list<std::pair<unsigned int, std::shared_ptr<Request>>> receive_reqs;
+  std::list<std::pair<unsigned int, std::shared_ptr<Request>>> receive_requests;
   auto current_request = std::make_shared<Request>();
 
   // Storage for the incoming data
@@ -237,7 +237,7 @@ push_parallel_nbx_helper(const Communicator & comm,
     if (possibly_receive_functor(
             comm, current_src_proc, *current_incoming_data, *current_request, tag))
     {
-      receive_reqs.emplace_back(current_src_proc, current_request);
+      receive_requests.emplace_back(current_src_proc, current_request);
       current_request = std::make_shared<Request>();
 
       // current_src_proc will now hold the src pid for this receive
@@ -246,7 +246,7 @@ push_parallel_nbx_helper(const Communicator & comm,
     }
 
     // Clean up outstanding receive requests
-    receive_reqs.remove_if([&act_on_data, &incoming_data](std::pair<unsigned int, std::shared_ptr<Request>> & pid_req_pair)
+    receive_requests.remove_if([&act_on_data, &incoming_data](std::pair<unsigned int, std::shared_ptr<Request>> & pid_req_pair)
                            {
                              auto & pid = pid_req_pair.first;
                              auto & req = pid_req_pair.second;
@@ -273,7 +273,7 @@ push_parallel_nbx_helper(const Communicator & comm,
                              return false;
                            });
 
-    reqs.remove_if([](Request & req)
+    requests.remove_if([](Request & req)
                    {
                      if (req.test())
                      {
@@ -289,7 +289,7 @@ push_parallel_nbx_helper(const Communicator & comm,
 
 
     // See if all of the sends are finished
-    if (reqs.empty())
+    if (requests.empty())
       sends_complete = true;
 
     // If they've all completed then we can start the barrier
@@ -300,7 +300,7 @@ push_parallel_nbx_helper(const Communicator & comm,
     }
 
     // Must fully receive everything before being allowed to move on!
-    if (receive_reqs.empty())
+    if (receive_requests.empty())
       // See if all proessors have finished all sends (i.e. _done_!)
       if (started_barrier)
         if (barrier_request.test())
@@ -496,7 +496,7 @@ void pull_parallel_vector_data(const Communicator & comm,
 
   // First index: order of creation, irrelevant
   std::vector<std::vector<std::vector<datum,A>>> response_data;
-  std::vector<Request> response_reqs;
+  std::vector<Request> response_requests;
 
   // We'll grab a tag so we can overlap request sends and receives
   // without confusing one for the other
@@ -504,7 +504,7 @@ void pull_parallel_vector_data(const Communicator & comm,
 
   auto gather_functor =
     [&comm, &gather_data, &act_on_data,
-     &response_data, &response_reqs, &tag]
+     &response_data, &response_requests, &tag]
     (processor_id_type pid, query_type query)
     {
       std::vector<std::vector<datum,A>> response;
@@ -521,7 +521,7 @@ void pull_parallel_vector_data(const Communicator & comm,
         {
           Request sendreq;
           comm.send(pid, response, sendreq, tag);
-          response_reqs.push_back(sendreq);
+          response_requests.push_back(sendreq);
           response_data.push_back(std::move(response));
         }
     };
@@ -536,7 +536,7 @@ void pull_parallel_vector_data(const Communicator & comm,
   // non-blocking APIs with this data type.
   //
   // FIXME - implement Derek's API from #1684, switch to that!
-  std::vector<Request> receive_reqs;
+  std::vector<Request> receive_requests;
   std::vector<processor_id_type> receive_procids;
   for (std::size_t i = 0,
        n_queries = queries.size() - queries.count(comm.rank());
@@ -555,7 +555,7 @@ void pull_parallel_vector_data(const Communicator & comm,
       act_on_data(proc_id, querydata, received_data);
     }
 
-  wait(response_reqs);
+  wait(response_requests);
 }
 
 } // namespace TIMPI
--- a/src/algorithms/include/timpi/parallel_sync.h
+++ b/src/algorithms/include/timpi/parallel_sync.h
@@ -52,14 +52,24 @@ namespace TIMPI {
  * All receives and actions are completed before this function
  * returns.
  *
- * Note: it is very important that the message tag be completely
- * unique to each invocation
+ * Two methods exist: one with a const reference to the data and another
+ * with an rvalue reference to the data. If you wish to use move semantics
+ * within the data received in \p act_on_data, use the method that passes
+ * the data as an rvalue reference.
  */
+ ///@{
 template <typename MapToVectors,
           typename ActionFunctor>
 void push_parallel_vector_data(const Communicator & comm,
                                const MapToVectors & data,
                                const ActionFunctor & act_on_data);
+
+template <typename MapToVectors,
+         typename ActionFunctor>
+void push_parallel_vector_data(const Communicator & comm,
+                               MapToVectors && data,
+                               const ActionFunctor & act_on_data);
+ ///@}
 
 /**
  * Send query vectors, receive and answer them with vectors of data,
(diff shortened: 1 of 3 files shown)


loganharbour:
This is the first attempt. Documentation isn't done yet and I haven't added in the move-capable methods. Still trying to figure out if I can generalize those further.

src/algorithms/include/timpi/parallel_sync.h
```diff
@@ -36,6 +36,158 @@
...
+template <typename MapToContainers,
+          typename SendFunctor,
+          typename PossiblyReceiveFunctor,
+          typename ActionFunctor>
+void
+push_parallel_nbx_helper(const Communicator & comm,
+                         MapToContainers & data,
+                         SendFunctor & send_functor,
+                         PossiblyReceiveFunctor & possibly_receive_functor,
+                         ActionFunctor & act_on_data)
```
loganharbour, line 48: @roystgnr - when you look at this, I need ideas on how to get this working with the rvalue `data`

roystgnr:
Cleaned up and force-pushed ... and yet I'm still struggling to figure out how to get an rvalue version working without just cut-and-pasting a ton of code.

All the usual idioms are for cases where the function itself needs to own its argument - so the C++11 thing to do is something like pass-by-value, and then the argument gets move-constructed if you have an lvalue or copy-constructed if you have an rvalue.

But here we want to move when given an rvalue, so we can handle containers of unique_ptr, but *not* copy when we get an lvalue, because we have no need to and some of our containers are too huge to copy cheaply.  I'm not seeing a common idiom to do that without a ton of code duplication.

src/algorithms/include/timpi/parallel_sync.h
```diff
@@ -153,34 +153,29 @@ void pull_parallel_vector_data(const Communicator & comm,
...
           typename ActionFunctor>
-void push_parallel_vector_data(const Communicator & comm,
-                               const MapToVectors & data,
-                               const ActionFunctor & act_on_data)
+void
+push_parallel_nbx_helper(const Communicator & comm,
+                         const MapToContainers & data,
+                         SendFunctor & send_functor,
+                         PossiblyReceiveFunctor & possibly_receive_functor,
+                         ActionFunctor & act_on_data)
```
lindsayad, line 165: Per offline discussion I would vote to make functors `const` until the compiler proves me otherwise

lindsayad (changes requested):

src/algorithms/include/timpi/parallel_sync.h
loganharbour: Ah, yeah. That should actually be const.

This was a leftover from when I was screwing around with using the helper for both the move method and the const method.

loganharbour:
Okay, this is ready for review. There are methods with move semantics for both `push_parallel_vector_data()` and `push_parallel_packed_range()`, with what I think is good code minimization.

I'd appreciate your eye too @lindsayad

test/packed_range_unit.C
```diff
@@ -31,6 +33,32 @@ struct null_output_iterator
...
+  static unsigned int packed_size(typename std::vector<int>::const_iterator) { return 1; }
+  static unsigned int packable_size(const std::unique_ptr<int> &, const void *) { return 1; }
+
+  template <typename Iter>
+  static void pack(const std::unique_ptr<int> & object, Iter data_out, const void *) { data_out = *object; };
+
+  template <typename BufferIter>
+  static std::unique_ptr<int> unpack(BufferIter in, const void *)
+    {
+      return std::make_unique<int>(*in++);
```
loganharbour, line 55: May need to get something `libmesh_make_unique` esque for this

roystgnr: We can wrap this in a test like
```
#if __cplusplus > 201402L
```
if you prefer.  We don't want to require any C++14 code in TIMPI itself yet, but I'd be happy to require it in the test suite.

roystgnr: (in a subset of the test suite, I mean - I of course at least want tests that don't require C++14 to run with C++11...)

src/algorithms/include/timpi/parallel_sync.h
```diff
@@ -357,6 +418,27 @@ void push_parallel_vector_data(const Communicator & comm,
...
+
+
+template <typename MapToVectors,
+          typename ActionFunctor>
+void push_parallel_vector_data(const Communicator & comm,
+                               MapToVectors && data,
+                               const ActionFunctor & act_on_data)
+{
+  auto moved_data = std::move(data);
+  push_parallel_vector_data_helper(comm, moved_data, act_on_data);
```
roystgnr, line 438: Is this really correct?  This looks like the "move 'v' into 'vp'; copy 'vp' into 'w'" case from https://stackoverflow.com/questions/8556148/whats-the-type-of-std-move

loganharbour: Yeah, you're definitely right here. Changed to:

```
auto moved_data(std::move(data));
```

test/packed_range_unit.C
loganharbour: Wrapped all of the checks with the packed unique_ptr with the above.

loganharbour:
Wonderful! Great success. I think I'm good with this where it is.

src/algorithms/include/timpi/parallel_sync.h
roystgnr: Does that do anything different, though?  I thought the catch here was that, whereas std::move returns a temporary (which thus gets passed as an rvalue), once you initialize any named variable from it, *that named variable* is no longer a temporary and (unless it is itself wrapped in std::move) it will only get passed as an lvalue.

roystgnr: Okay - this is actually correct in context, but that's because the code it's calling is incorrect, blindly moving its argument regardless of whether the argument should be moved (as in this call) or shouldn't (in the const reference API).  I've uploaded a new unit test that demonstrates the problem.

lindsayad:
This PR caused be to go down a bit of a hole to really think about what `const` means when related to closure objects. So I'm posting this code for my own edification. You guys may have already know all this, but I did not
```c++
#include <iostream>
#include <memory>
#include <vector>

class A {
public:
  void foo() {}

  void bar() const {}
};

// Per https://en.cppreference.com/w/cpp/language/lambda there is only ever one
// operator() method defined in a closure. If a lambda is marked as mutable, then
// operator() will be non-const...unless it seems as shown below it doesn't capture
// any data by value in which case operator() will be const. If the lambda is not marked
// mutable then operator() is always const

template <typename ActionFunctor, typename DataType>
void push_parallel_packed_range(const ActionFunctor & act_on_data, DataType & data)
{
  std::cout << "const overload\n";
  act_on_data(data);
}

/**
 * Require this overload in order to compile
 *     push_parallel_packed_range(mutable_lamb, from);
 * Without it we get this error with g++:
 * const-functor.cpp: In instantiation of ‘void push_parallel_packed_range(const ActionFunctor&, DataType&)
 * [with ActionFunctor = main()::<lambda(std::vector<std::unique_ptr<A> >&)>; DataType = std::vector<std::unique_ptr<A> >]’:
 * const-functor.cpp:60:48:   required from here
 * const-functor.cpp:15:14: error: no match for call to
 * ‘(const main()::<lambda(std::vector<std::unique_ptr<A> >&)>) (std::vector<std::unique_ptr<A> >&)’
 * 15 |   act_on_data(data);
 *    |   ~~~~~~~~~~~^~~~~~
 * const-functor.cpp:54:23: note: candidate: ‘main()::<lambda(std::vector<std::unique_ptr<A> >&)> mutable’ (near match)
 * 54 |   auto mutable_lamb = [a](std::vector<std::unique_ptr<A>> & arg) mutable
 *    |                       ^
 * const-functor.cpp:54:23: note:   passing ‘const main()::<lambda(std::vector<std::unique_ptr<A> >&)>*’ as ‘this’
 * argument discards qualifiers
 *
 * TLDR: the mutable_lamb closure object will only define non-const operator() so if the closure object itself (or reference to it)
 * is marked const, then there is no viable method to call
 */
template <typename ActionFunctor, typename DataType>
void push_parallel_packed_range(ActionFunctor & act_on_data, DataType & data)
{
  std::cout << "non-const overload\n";
  act_on_data(data);
}

int main()
{
  std::vector<std::unique_ptr<A>> to, from;
  from.emplace_back(std::make_unique<A>());

  A a, b;
  const A c;

  // Not marked mutable so operator() const is the method created in the closure
  const auto my_lamb = [a, &b, &c, &to](std::vector<std::unique_ptr<A>> & arg)
                   {
                     for (auto & value : arg)
                       to.emplace_back(std::move(value));

                     // Below causes program not to compile
                     // Unless the lambda is specified mutable, things captured by copy are const-qualified
                     // a.foo();

                     b.foo();
                     a.bar();

                     // Below causes program not to compile.
                     // c is const in the lambda enclosing scope
                     // c.foo();
                   };

  // We can mark this as const and the const ActionFunctor & push_parallel_packed_range overload
  // works fine because even though we've marked this as mutable, we're not capturing any data
  // by value and so the operator() const method is the one present in the closure
  // (as opposed to a non-const operator())
  const auto mutable_no_captures = [](std::vector<std::unique_ptr<A>> & arg) mutable
                               {
                               };

  // If the below lambda is marked const, then the const ActionFunctor & push_parallel_packed_range
  // overload is selected and then compilation fails because there is no operator() const method in the
  // closure because we've marked this lambda mutable and we are capturing data by value
  /*const*/ auto mutable_lamb = [a](std::vector<std::unique_ptr<A>> & arg) mutable
                   {
                     a.foo();
                   };

  push_parallel_packed_range(my_lamb, from);
  push_parallel_packed_range(mutable_no_captures, from);
  push_parallel_packed_range(mutable_lamb, from);

  // A general note: if we didn't mark my_lamb or mutable_lamb as const, then the non-const push_parallel_packed_range
  // overload is the overload that gets called, but everything is hunky-dory because even though the lambda closures
  // operator() method is const, non-const objects can of course call const methods
}
```

lindsayad (commented):
So a little pedantic, but when choosing whether to mark our functors as `const` or not, we will be deciding whether we want to allow our code to work with non-const `operator()`. In the case of lambdas it means deciding whether we want to allow users to mutate data captured by value. Despite what I told @loganharbour earlier on Slack, I don't see much of a reason to restrict our users here. If they know enough to mark their lambdas as mutable, then I think I trust them not to do stupid stuff in their lambda bodies, or if they do it's on their own head. So I'm game to remove the `const` qualification from all of our functors.

lindsayad:
Also, @roystgnr did you have a motivation behind passing functors by reference as opposed to by value?

src/algorithms/include/timpi/parallel_sync.h
```diff
@@ -153,34 +158,29 @@ void pull_parallel_vector_data(const Communicator & comm,
...
           typename ActionFunctor>
-void push_parallel_vector_data(const Communicator & comm,
-                               const MapToVectors & data,
-                               const ActionFunctor & act_on_data)
+void
+push_parallel_nbx_helper(const Communicator & comm,
+                         MapToContainers & data,
+                         SendFunctor & send_functor,
+                         PossiblyReceiveFunctor & possibly_receive_functor,
+                         const ActionFunctor & act_on_data)
```
lindsayad, line 170: Should this ever directly be used by a library user? If not can we put this in a `detail` namespace or something?

roystgnr: IMHO no, and I'll do that now.

lindsayad:
> And if you want to pass by reference for efficiency, then you want to pass by const reference so that temporaries are allowed.

I totally agree with this. I think we either do `const Functor &` or `Functor`.

> Should we change this? I could be talked into it; the STL uses pass-by-value (because of C++03 limitations, though, IIRC) so C++ programmers ought to be relatively used to writing cheap-to-copy functors.

But to be consistent with STL I would lean towards pass-by-value. Who knows maybe someday TIMPI will be the parallel STL... :smile:

lindsayad:
But my opinion is not strong. I definitely see the argument for `const Functor &` in the case that our users have a lot of state data. I don't really think they should be doing that, but I don't know if I want to make it prohibitive either

roystgnr:
I think all that has me worried about the change is that it's technically API-breaking.  My inclination would be to save it for when we try out multithreading, because I fear that that would entail API changes as well, and if that's the case then it would be nice to just require one round of userspace changes instead of two.

lindsayad:
That's fine with me

src/algorithms/include/timpi/parallel_sync.h
lindsayad: Now that we've concluded the discussion below, let's change `send_functor` and `possibly_receive_functor` to be `const Functor &`

roystgnr:
In that case I think I'm happy with the current state of this, but we'll wait for @loganharbour to look over the final state before we merge.

src/algorithms/include/timpi/parallel_sync.h
```diff
@@ -189,290 +192,217 @@ void push_parallel_vector_data(const Communicator & comm,
...
+      // Look for data from anywhere
+      current_src_proc = any_source;
 
+      // Check if there is a message and start receiving it
+      if (possibly_receive_functor(comm, current_src_proc,
+                                   *current_incoming_data,
+                                   current_request, tag))
+        {
+          receive_requests.emplace_back(current_src_proc,
+                                        std::move(current_request));
```
loganharbour, line 250: @roystgnr did you conclude that we were good with moving the request?

loganharbour:
I'm happy with this and it's working for my use case (the unique_ptr move).

src/algorithms/include/timpi/parallel_sync.h
roystgnr: Yeah; we don't have a move constructor right now so it doesn't actually do anything, but we might add one and when we do it'd be nice for it to just automatically get employed here.

roystgnr:
I think ed3c3af probably isn't necessary, because of the internal reference counting Request does, but I can't fault anyone for trusting std:: code more than roy:: code.

Issue #69, opened 2021-05-29
Support for maps of more complex types in set_union


bwspenc:
I'd like to be able to pass a map of more complex data types to `communicator::set_union()`. Specifically, I'm interested in passing in the following type from a MOOSE code:

`std::map<unsigned int, std::vector<RealVectorValue>>`

It looks like currently there is no support for `std::map<unsigned int, std::vector<anything>>`.

bwspenc:
@austinltam

loganharbour:
```
  std::map<unsigned int, std::vector<RealVectorValue>> some_map;
  comm().set_union(some_map);
```

Compiles for me no problem. Did you include `parallel_sync.h` ?

bwspenc:
I don't have ready access to the specific code, but @austinltam will look into whether that header include was missing this evening. We actually spent a while looking into this with @roystgnr. Hopefully it's as easy as including another header!

austinltam:
I included parallel_sync.h but I am still getting the buffer_type error as shown below.

 error: no type named 'buffer_type' in 'struct libMesh::Parallel::PairBufferTypeHelper<unsigned int, false, std::vector<libMesh::VectorValue<double>, std::allocator<libMesh::VectorValue<double> > >, false>'
  138 |                                         Has_buffer_type<Packing<T2>>::value>::buffer_type buffer_type;

roystgnr:
Try adding the macro invocation in https://github.com/libMesh/TIMPI/pull/70/commits/3141b802a26b07afcadff2db3d44f2d9a14ca117

and see if that works for you?

roystgnr:
Ping?  I finally got around to fixing the name collision in https://github.com/libMesh/TIMPI/pull/70 and I'd like to start getting it moved into libMesh and Moose if it works for you.  I have vacation coming up soon and I'm hoping to get everything I've been working on this month pushed downstream first, with enough advance notice to fix any problems that occur.

bwspenc:
@roystgnr Sorry it took me so long to get to this!  I just had a chance to test the code in #70, and verified that it works in our MOOSE code. I tested it by doing the following:

1) I added the following code to the beginning of the XFEMMaterialStateMarkerBase::finalize() method (in moose/modules/xfem/src/userobjects), which is where the code is that we want this feature for:
```
  unsigned int pid = processor_id();
  std::cout<<"BWS0 pid: "<<pid<<std::endl;
  RealVectorValue rvv;
[...]

PR #70, opened 2021-06-02
Packing<vector> support, set_union unit tests

roystgnr:
This is hopefully sufficient to solve the issue in #69, and hopefully won't break (or pessimize, if our custom vector<vector> overloads were to suddenly stop getting preferred to packed_range dispatch) any other code paths, but I'm not 100% certain of any of that.


bwspenc:
As I noted in issue #69, I tested this and verified that it works correctly on my data structure of interest.

PR #71, opened 2021-07-25
Add override keyword

pbauman:
This silences a warning I was getting from GCC's -Wsuggest-override.


roystgnr (approved):
Same questions as in the libMesh ticket. ... except I don't think we've moved our paranoid-warnings support to the autoconf submodule yet, so we'll still be stuck getting any test coverage on that indirectly from libMesh submodule updates.

PR #76, opened 2021-11-03
Packing size fixes

roystgnr:
Pinging @loganharbour, who managed to trigger the error that this PR fixes for me.  I think I have a decent unit test here, but do you perhaps still got your pre-workaround code handy, so we can try it directly on that too?


roystgnr:
Why am I just fixing this for std::pair when we have a dozen other containers that have the same problem?  I'll tack on fixes and unit tests for those to this same PR.

PR #77, opened 2021-11-04
autoconf-submodule update, to support more compiler flags

roystgnr:
This gets --enable-paranoid-warnings and --enable-werror working, so we can turn those on in all our CI tests.  At least on my system the code is already passing, but this might be helpful in catching future errors while they're still TIMPI PRs and not libMesh TIMPI submodule updates.

Sorry, @loganharbour, didn't make it to --enable-coverage yet.


roystgnr:
I have to cherry-pick that whitespace fix into every PR until one gets merged, don't I?

loganharbour:
> I have to cherry-pick that whitespace fix into every PR until one gets merged, don't I? 

Yep :)

PR #78, opened 2021-11-09
Make DataType destructor virtual

lindsayad:
As evidenced by the number of errors I get if I make this destructor
protected, we delete from `DataType` quite a bit. Consequently we have
to make sure that the `StandardType` destructor gets called because that
is generally what calls `free()`. This is ultimately what slew
idaholab/moose#19246, causing valgrind errors in
https://civet.inl.gov/job/880361/, and leading to reversion in
idaholab/moose#19331

### DIFF
--- a/src/parallel/include/timpi/data_type.h
+++ b/src/parallel/include/timpi/data_type.h
@@ -68,7 +68,7 @@ public:
     this->commit();
   }
 
-  ~DataType () = default;
+  virtual ~DataType () = default;
 
   DataType & operator = (const DataType & other) = default;
   DataType & operator = (DataType && other) = default;


roystgnr (approved):

jwpeterson:
This seems like something important enough to backport to the TIMPI in the in-progress release? If so I could look into what that would entail, probably a new bootstrapped TIMPI tag that we could then update the release submodule to...

roystgnr:
Yeah, this is pretty embarrassing, probably worth backporting.  I'm just not sure how we never managed to trigger it before, it took a complicated combination of nested MetaPhysicL classes to do so ... but in *theory* even a plain std::pair could have broken in the same way.

Let's hold off on the backport for now, though.  The problem here is just a leak, allocations of derived StandardType classes that aren't freed.  While looking into this, though, I noticed that we might have a similar but more serious issue that could lead to dangling references to derived user-defined MPI types.  I'm hoping to have a fix for that (or an announcement that it was just a false alarm) soon.

lindsayad:
Is this the copy assignment operator concern?

roystgnr:
Yeah; in hindsight I'll bet that TIMPI code is actually doing RVO and so not even triggering the bad copy assignment, and most (perhaps all?) of our other copies are copy construction rather than copy assignment, but it just feels like it might be an accident waiting to happen.

lindsayad:
Yea I think it's good if we follow the Rule of 3

PR #79, opened 2021-11-10
Implement copy assignment operators for StandardType

lindsayad:
Maybe I can drum up a test that would not have been valgrind clean before this change. Shouldn't be very hard

### DIFF
--- a/src/parallel/include/timpi/standard_type.h
+++ b/src/parallel/include/timpi/standard_type.h
@@ -195,6 +195,13 @@ TIMPI_STANDARD_TYPE(long double,MPI_LONG_DOUBLE);
       timpi_call_mpi (MPI_Type_dup (t._datatype, &_datatype));
     }
 
+    StandardType & operator=(StandardType & t)
+    {
+      this->free();
+      timpi_call_mpi(MPI_Type_dup(t._datatype, &_datatype));
+      return *this;
+    }
+
     ~StandardType() {
       this->free();
     }
@@ -276,6 +283,13 @@ public:
       (MPI_Type_dup (t._datatype, &_datatype));
   }
 
+  StandardType & operator=(StandardType & t)
+  {
+    this->free();
+    timpi_call_mpi(MPI_Type_dup(t._datatype, &_datatype));
+    return *this;
+  }
+
   ~StandardType() { this->free(); }
 
   static const bool is_fixed_type = true;
@@ -350,6 +364,13 @@ public:
       (MPI_Type_dup (t._datatype, &_datatype));
   }
 
+  StandardType & operator=(StandardType & t)
+  {
+    this->free();
+    timpi_call_mpi(MPI_Type_dup(t._datatype, &_datatype));
+    return *this;
+  }
+
   ~StandardType() { this->free(); }
 
   static const bool is_fixed_type = true;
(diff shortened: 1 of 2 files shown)


src/parallel/include/timpi/standard_type.h
```diff
@@ -386,7 +407,7 @@ void BuildStandardTypeVector<n_minus_i>::build
     ith_type;
 
   out_vec.emplace_back
-    (new StandardType<ith_type>
+    (std::make_unique<StandardType<ith_type>>
```
roystgnr, line 410: How'd you catch this?  We were doing it wrong before, technically, but the wrongness was some subtle "if vector resize can't allocate memory and throws and we catch the exception then we leak the new'd memory" thing, not anything I'd have expected the new valgrind tests to pick up.

roystgnr (approved):
I'd love to have a new test that would have triggered the bug fixed here, but I think I'm content merely to see that the fix isn't breaking any tests.

src/parallel/include/timpi/standard_type.h
lindsayad: This was when I was exploring what *might* be causing the leaks.

lindsayad:
Ok with the test I just added I didn't even have a chance to get a valgrind error without the assignment operator fix. Instead I got this from my MPI implementation:
```
Assertion failed in file src/mpi/datatype/type_free.c at line 38: (((datatype_ptr)))->ref_count >= 0
0x7ffff7d8c707 ???
	???:0
0x7ffff7ca889f ???
	???:0
0x7ffff7ca8cea ???
	???:0
0x55555556700a _ZN5TIMPI8DataType4freeEv
	../../src/parallel/include/timpi/data_type.h:99
0x55555557623b _ZN5TIMPI12StandardTypeISt4pairIjjEvED2Ev
	../../src/parallel/include/timpi/standard_type.h:279
0x5555555660b5 _Z26testStandardTypeAssignmentv
	../../test/parallel_unit.C:856
0x5555555666ff main
	../../test/parallel_unit.C:913
0x7ffff78680b2 ???
	???:0
0x55555555cf3d ???
	???:0
0xffffffffffffffff ???
	???:0
internal ABORT - process 0
```

src/parallel/include/timpi/standard_type.h
roystgnr: I'm an idiot.  "wrong before, technically" was a deliberate decision: std::make_unique isn't in C++11.

And because the current default libMesh configuration is "ask for C++17 or C++14 before settling for C++11", we didn't even catch this in the "Minimum $CXX version" CI tests in https://github.com/libMesh/libmesh/pull/3082

I'm inclined to just let it slide, though, since "update our required C++ version to C++17 by the end of the year" is still a goal of mine and since we're almost all out of year.

lindsayad: oopsie. I've only been living in required c++14 land for a few months in MOOSE and I've already stopped thinking about being c++11 compliant

roystgnr: Now that I think about it, in fact, the original plan was "One more C++11 compatible release and then we upgrade".  We haven't finalized the release yet but we have branched; it should be okay to push newer C++ to master.

PR #80, opened 2021-11-11
Expanded OpFunction capabilities

roystgnr:
This solves the tiny leak in our quad-precision on-the-the-fly-`MPI_Op` support, and more importantly it extends that support to std::pair and adds a unit test for it.  There's still a ton of work to be done here (and in MetaPhysicL), but I'm throwing this up now that it's passing tests, to see if anyone (or any CI tests) has criticisms.

This is our first-ever "do not merge"-labeled TIMPI PR because I'm still not 100% sure of the API myself.  I think I may move the new static members and APIs from `TIMPIInit` to `SemiPermanent` (even though the `TIMPIInit` destructor will still have to access them, we can make it a friend class), and I should probably replace `FreeOp` with something that does both halves of the create+free, to make that be a proper RAII class.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -106,6 +106,7 @@ include_HEADERS += parallel/include/timpi/status.h
 
 # utilities
 include_HEADERS += utilities/include/timpi/ignore_warnings.h
+include_HEADERS += utilities/include/timpi/semipermanent.h
 include_HEADERS += utilities/include/timpi/timpi.h
 include_HEADERS += utilities/include/timpi/timpi_assert.h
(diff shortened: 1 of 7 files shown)


roystgnr:
@jwpeterson noticed that I let some C++14 slip in here.  I then noticed that we already let C++14 slip into libMesh via https://github.com/libMesh/TIMPI/pull/79 and https://github.com/libMesh/libmesh/pull/3082 ... but since we've already branched our planned "final C++11-compatible release" in libMesh, we should be fine to start adding newer C++ now.  Need to be careful when we're backporting bugfixes, though...

PR #82, opened 2021-11-16
Update compiler requirements from C++11 to C++14

roystgnr:
This should keep us working with older compilers (once that support
C++14 but don't default to it) for a little longer, until we're ready for C++17.


roystgnr:
Merging right away so we can kick the tests in https://github.com/libMesh/MetaPhysicL/pull/5

PR #83, opened 2021-11-17
Code coverage support

roystgnr:
This is akin to what's in libMesh, except we're not going to bother with a redundant copy of libMesh's contrib/lcov/ - even for non-libMesh users that's a standard package now in most OS repositories.


Makefile.am
```diff
@@ -28,6 +28,42 @@ test_headers:
...
+# distribution like $LIBMESH_DIR/contrib/lcov/lconv-1.15/bin/
+
+lcov-report:
+	@mkdir -p $(lcov_dir)
+	lcov --compat-libtool --directory . --capture --output-file $(lcov_dir)/lcov.info
+	lcov --list-full-path -l $(lcov_dir)/lcov.info | grep -v "`cd -P $(top_srcdir)/src && pwd`" | cut -d\| -f1 > $(lcov_dir)/remove
+	lcov -q -r $(lcov_dir)/lcov.info `cat $(lcov_dir)/remove` > $(lcov_dir)/lcov.cleaned.info
+	@rm $(lcov_dir)/remove
+	@mv -f $(lcov_dir)/lcov.cleaned.info $(lcov_dir)/lcov.info
+	genhtml -t "libMesh" -o $(lcov_dir) $(lcov_dir)/lcov.info
```
loganharbour, line 53:
```suggestion
	genhtml -t "TIMPI" -o $(lcov_dir) $(lcov_dir)/lcov.info
```

configure.ac
```diff
@@ -85,6 +85,14 @@ ACSM_CXX_COMPILER_STANDARD([2014], [2017])
 #-----------------------------------------------------
 TIMPI_SET_CXX_FLAGS
 
+#------------------------------------------------------
+# Checks for code coverage
+# this must be done after we look for our own compilers
+#------------------------------------------------------
+ACSM_CODE_COVERAGE
+timpi_CFLAGS="$GCOV_FLAGS $timpi_CFLAGS"
+timpi_CXXFLAGS="$GCOV_FLAGS $timpi_CXXFLAGS"
```
loganharbour, line 94: This is actually all that we need for CIVET right? Just the addition of the flags?

loganharbour (changes requested):
Right now in CIVET, we run that extra script that replaces filenames in the coverage report in the build directory with their actual location in the source.

should we move that capability here?

configure.ac
roystgnr: Yeah, in theory.  I just figured it'll be easier to test and easier for other users if we have some make targets too.

roystgnr:
Which script is that?  The "Replace the includes" bit in scripts/coverage_extract_and_store_libmesh.sh?

roystgnr:
If so: is that necessary here?  With libMesh we have the weirdness where we store files like `include/some_category/foo.h` but then we include them via symlinks like `include/libmesh/foo.h` to keep the namespace unpolluted.  With TIMPI we store files directly as `src/some_category/include/timpi/foo.h`, no need for symlinks, and as best as I can tell from a `make coverage` run no need for postprocessing either.

configure.ac
loganharbour: Yep - works for me

loganharbour:
> If so: is that necessary here? 

Not necessarily. Just wanted to bring it up if you had opinions.

loganharbour (approved):

PR #86, opened 2022-01-06
Simplify logic in parallel sync nbx helper

loganharbour:
Random improvements while I'm digging through this.

Will probably need `timpi_make_unique` for c++ 11

### DIFF
--- a/src/algorithms/include/timpi/parallel_sync.h
+++ b/src/algorithms/include/timpi/parallel_sync.h
@@ -28,7 +28,6 @@
 #include <iterator>    // inserter
 #include <list>
 #include <map>         // map, multimap, pair
-#include <memory>      // shared_ptr, make_shared
 #include <type_traits> // remove_reference, remove_const
 #include <utility>     // move
 #include <vector>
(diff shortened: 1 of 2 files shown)


roystgnr:
We already upgraded TIMPI's and libMesh's requirements to C++14.  Not in a proud way, mind you, but rather because we'd already let C++14 code slip into TIMPI without catching it...

roystgnr:
Huh.  This seems reasonable to me, but there must have been some reason why @friedmud used `shared_ptr` rather than `unique_ptr` in the first place.

You tried to change from `shared_ptr<Request>` to plain `Request` at one point and had to revert that, right?  What happened there?

loganharbour:
The issue was we need move semantics because we don’t wanna copy construct the requests.

roystgnr:
I thought it was something like that.  But what confused me is that we do try to make our Request class have shared_ptr type semantics internally.  So there has to be something going wrong with that.

loganharbour:
That’s a good point. I’ll look into it

loganharbour:
I looked into this some. We could probably get away with using direct objects and calling `std::move()` on them, but it doesn't look like there is even a well defined state of an object that is moved from... so I think that using a unique_ptr is still the correct choice here.

Even with `std::vector` (see https://stackoverflow.com/questions/17730689/is-a-moved-from-vector-always-empty) doesn't have a standard defined state after a move. We could define `&&` constructors for `Request`, but I think that's excessive.

I looked into doing something like (for a vector)

```
container_type current_incoming_data;
auto new_data = std::move(current_incoming_data);
// do something with current_incoming_data, or current_incoming_data.clear()
// but just ended up with logic errors
```

If we can't even do this with a `std::vector` in a clean and obvious manner, I don't think it's worth getting to work with a `Request`. The code as written is clear and concise.

roystgnr:
The Request behavior is a separate possible (probable?) bug, so not a show-stopper for merging this.  This does look like a clear improvement to me.  But the situation also screams "Chesterton's Fence" so I'd rather not merge until Derek has a chance to chime in, which probably won't be soon.

roystgnr (approved):
Okay, I'm really loving this version.  Can you test it manually with libMesh and Moose before we merge, though?

loganharbour:
Good to go.

friedmud:
Hmmm - this is ok I guess... I don't like that semantically having 1 thing in the buffer is the "empty" case... but I get it.  I guess the improvement here is not using `shared_ptr`?

roystgnr:
Not using shared_ptr, and no longer failing oversized pulls.

> I don't like that semantically having 1 thing in the buffer is the "empty" case...

It's like we just reinvented C-style strings, isn't it?  I couldn't think of a better solution, though.

PR #89, opened 2022-01-25
Add Intel(R) icc 21 support

mkhon:
Requires libMesh/autoconf-submodule#6


roystgnr:
Well, not quite *requires* - the feature will be non-functional until we update the submodule but at least it won't regress anything in the meantime.

roystgnr (approved):

PR #90, opened 2022-02-15
Do our own checks before memcpy

roystgnr:
When we communicate variable-size containers of fixed-size classes, this
will require us to be a little strict about those classes' constructors
... but not as impossibly strict as the new gcc warnings would have us
be.

Hopefully this will fix (in combination with some libMesh constructor cleanup)
https://github.com/libMesh/libmesh/issues/3157


roystgnr:
This works in my libMesh tests, so should be good to merge and to turn into 1.8.3.

PR #91, opened 2022-02-23
Remove static_assert(is_trivially_copyable)

roystgnr:
This was enough for our own classes, but it still gives false positives
for C++ standard classes.  Testing for a StandardType should have been
enough here.

Refs #90


jwpeterson:
See also: discussion on libMesh/libmesh#3157

PR #95, opened 2022-03-29
Better fixes for Packing specializations

roystgnr:
My earlier fixes for strings broke sufficiently-long non-strings.

The new tests here cover this bug; I'm also going to run some test privately to see if I can trigger any downstream regressions and to ensure that the fix here also works for the sort of bug that we saw originally.


roystgnr:
No, this isn't quite there.  I'll take a crack at the remaining issues tomorrow.

roystgnr:
This seems to be working for me downstream too now.

PR #100, opened 2023-01-16
Packing<string> should use unsigned int buffer_type too

roystgnr:
I wanted to mix and match buffer types; I failed; I admit that now.

This fixes mixed types like `pair<int,string>`


loganharbour:
🥳

PR #106, opened 2023-01-31
set_union fixes

### LINKED ISSUE #105: Issues with set_union()
I ran into some issues with set_union() for the following container types:
```
std::map<unsigned int, std::set<boundary_id_type>> mapset;
std::map<unsigned int, std::map<boundary_id_type, Real>> mapmap;
```
The code compiles fine but either does not do the `set_union()` correctly (mapset) or crashes (mapmap). I am pasting some standalone example codes in the comments below, GitHub won't let me attach them as files for some reason.

roystgnr:
Closes #105 IMHO - @jwpeterson should chime in on that too.

Important bug fix here, too.  If Peter programmer packs a padded pair into packets, how many packets did Peter programmer pack?

I'm just now realizing I haven't yet tested this with --disable-mpi.  Hopefully CI will tell me how bad a mistake that was.

### DIFF
--- a/src/parallel/include/timpi/communicator.h
+++ b/src/parallel/include/timpi/communicator.h
@@ -439,9 +439,13 @@ public:
   void sum(T & r, T & o, Request & req) const;
 
   /**
-   * Take a container of local variables on each processor, and
-   * collect their union over all processors, replacing the set on
-   * processor 0.
+   * Take a container (set, map, unordered_set, multimap, etc) of
+   * local variables on each processor, and collect their union over
+   * all processors, replacing the original on processor 0.
+   *
+   * If the \p data is a map or unordered_map and entries exist on
+   * different processors with the same key and different values, then
+   * the value with the lowest processor id takes precedence.
    */
   template <typename T>
   inline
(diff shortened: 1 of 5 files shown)


jwpeterson:
> I'm just now realizing I haven't yet tested this with --disable-mpi. Hopefully CI will tell me how bad a mistake that was.

Looks like "without MPI" tests are the only failures here.

jwpeterson (approved):
I'm a little worried that I will still call `set_union()` on types where it doesn't make sense, and then end up with runtime errors rather than a compile-time error or `static_assert` saying "this probably won't work like you expect". Definitely an improvement or what we had before though.

roystgnr:
That's a fair worry.  It's exactly the same worry as with `insert` or `merge`, though; so I fear the fix looks less like "make TIMPI safer" than like "make C++ safer".  They've been going in the right direction on that one; fingers crossed for C++79?

I'll merge this once I have the --disable-mpi fixes in ... but I was screwing up my builds for that and only *just* figured out why I wasn't replicating the problem, so it'll still take a bit.

PR #113, opened 2023-03-08
Carry rvalue references through the nbx helper

lindsayad:
We do so with the public APIs so let's do it with the helper too

### DIFF
--- a/src/algorithms/include/timpi/parallel_sync.h
+++ b/src/algorithms/include/timpi/parallel_sync.h
@@ -45,7 +45,7 @@ namespace TIMPI {
  * map entry; this will avoid any unnecessary communication.
  *
  * Data which is received from other processors will be operated on by
- * act_on_data(processor_id_type pid, const std::vector<datum> & data)
+ * act_on_data(processor_id_type pid, std::vector<datum> && data)
  *
  * If data exists for the local processor in the map, it will be acted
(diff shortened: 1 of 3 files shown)


loganharbour (approved):
👍🏼 looks good to me

not sure why we didn't do this previously...

loganharbour (changes requested):
Shouldn't we still test the reference case? I suspect folks have implemented the lambda as non auto and don't pass in an rvalue ref

lindsayad:
if the action functor takes the data as an lvalue reference then with the change here, their code fails to compile, hence why I had to change the unit tests here. But if their action functor takes the data as a const lvalue reference then their code will still work. In MOOSE when compiling the contact module, there was only one action functor I had to change, and it was in the framework `LowerDBlocksFromSidesetGenerator`

loganharbour (changes requested):
We should update the doxygen to reflect this; I was just working with @grmnptr and he was confused because the documentation explicitly says you get a const ref to the received containers. Not very intuitive when you're trying to use move semantics

loganharbour (approved):

lindsayad:
@roystgnr I'm going to wait until you approve before merging

roystgnr (approved):
I always hate having to change a test case or user code, but we really did never intend to support non-const lvalue references there.

PR #114, opened 2023-03-22
Don't define `Packing` specializations we can't pack

roystgnr:
A `container<SomethingUnpackable>` should itself be recognized as unpackable, even if other `container<T>` can be handled.

This should let user code test `Has_buffer_type<Packing<T>>::value` and see false for types we can't pack.

I ought to throw some `static_assert(!Has_buffer_type<Packing<Foo>>::value)` for a bunch of unpackable foo into the test suite before we merge this, but the headers should be in shape for Logan to play with.

### DIFF
--- a/src/parallel/include/timpi/packing.h
+++ b/src/parallel/include/timpi/packing.h
@@ -371,11 +371,13 @@ struct PairBufferTypeHelper<T1, false, T2, false>
 // specialization for std::pair
 template <typename T1, typename T2>
 class Packing<std::pair<T1, T2>,
-              typename std::enable_if<!TIMPI::StandardType<std::pair<T1, T2>>::is_fixed_type>::type>
+              typename std::enable_if<PairHasPacking<T1,T2>::value>::type>
 {
 public:
+  typedef typename std::remove_const<T1>::type cT1;
+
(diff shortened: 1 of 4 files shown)


loganharbour:
I like the static asserts. This makes sense to me.

Have you tried this with my problem case?

roystgnr:
Just tried it.  Either with only the changes to `canBroadcast<T>` or with your full branch, it's now compiling for me.

loganharbour (approved):
200w

src/parallel/include/timpi/packing.h
```diff
@@ -519,38 +521,72 @@ struct TupleBufferType<T1, MoreTypes...>
 template <typename Enable>
 class Packing<std::tuple<>, Enable> {};
 
-template <typename... Types>
-class Packing<std::tuple<Types...>,
-              typename std::enable_if<!TIMPI::StandardType<std::tuple<Types...>>::is_fixed_type>::type>
+/*
```
loganharbour, line 524: Don't forget to remove this

roystgnr: Whoops; of course

PR #115, opened 2023-04-07
Options to use older sync algorithms

roystgnr:
This might be interesting when benchmarking the new algorithms, and it might be useful when debugging weird parallel failures or MPI stacks, or it might only be useful because it adds test coverage for a bunch of previously uncovered overloads.

### DIFF
--- a/src/algorithms/include/timpi/parallel_sync.h
+++ b/src/algorithms/include/timpi/parallel_sync.h
@@ -377,6 +377,168 @@ push_parallel_nbx_helper(const Communicator & comm,
   const_cast<Communicator &>(comm).send_mode(old_send_mode);
 }
 
+template <typename MapToContainers,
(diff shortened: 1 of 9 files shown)


roystgnr:
After some thought I decided on a better name for one of the old options.  I have no idea what would be a better name for the other.  "ALLTOALL_COUNTS" is a confusing mouthful, but it at least captures the important point (and the weak point) of the algorithm.  So I'll merge this once it's done with CI again.

roystgnr:
> So I'll merge this once it's done with CI again.

Oh, semi-relevant: I've also been testing this branch with libMesh, trying to see if I can hit a use case that breaks one of the newer-older two options, with no failures so far.  Score one for TIMPI's internal test suite, which caught a lot of failures before I fixed them, albeit mostly in the N->M use cases.

PR #116, opened 2023-04-10
Avoid shadowing send()

roystgnr:
Newer compilers don't seem to complain about function parameters shadowing other function names, but gcc 7.5 doesn't like seeing "send" in one of our new overload implementations.  Let's just rename it everywhere.

### DIFF
--- a/src/parallel/include/timpi/communicator.h
+++ b/src/parallel/include/timpi/communicator.h
@@ -912,9 +912,9 @@ public:
                                     int>::type = 0>
   inline
   void send_receive(const unsigned int dest_processor_id,
-                    const T1 & send,
+                    const T1 & send_data,
                     const unsigned int source_processor_id,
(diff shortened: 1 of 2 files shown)


jwpeterson:
A great example of why people don't have `-Wshadow` enabled by default.

roystgnr:
Huh... the thing is, TIMPI tests *do* have `-Wshadow` enabled by default, including in a "Min gcc" recipe.

...

Aw hell.  The `timpi/Min_clang.cfg` still has `name = Min gcc` in it.  Somehow (a hash table it's overwriting the correct entry for?) that's what's eventually running under that recipe name.  And `git log -p` doesn't let me pin this one on Logan, either, that's my cut-and-paste-and-edit-but-not-enough bug.  Ah, well, I'll put up the fix and hopefully things like this get nipped in the bud next time.

PR #118, opened 2023-04-11
Add Communicator::sync_type setter via a string

loganharbour:
This better enables setting this sync type via parameters or cli args.


roystgnr (approved):

loganharbour:
Sounds good - we're good for the magic green button here then, please. I'll put the libmesh PR up.

PR #120, opened 2023-04-11
Add missing inline keyword

roystgnr:
This wasn't a problem when the function was only getting invoked via one object file's main(), but "duplicate symbol" becomes a problem when it's getting invoked via a dozen different compliation units in the same library downstream.


loganharbour (changes requested):
Should we

```
#ifndef NDEBUG
...
#endif
```

this method?

PR #121, opened 2023-04-17
NBX improvements

loganharbour:


roystgnr (approved):
This looks good.  Redundant if I understand the MPI standard and all MPI implementers do too, but I'm less and less convinced that any of us do ... and it's a cheap redundancy; I'm inclined to merge it even if it doesn't fix those sawtooth issues.

roystgnr:
> I'm inclined to merge it even if it doesn't fix those sawtooth issues.

In the sad light of "no it doesn't", I could go either way.  Maybe we just add that assertion, and don't bother with the redundant check in opt mode?

loganharbour:
Ok @roystgnr - ready

roystgnr (approved):

PR #123, opened 2023-04-24
Fix compile error with NVCC and OpenMPI

### LINKED ISSUE #122: Compile error with NVCC + OpenMPI
I'm seeing the following error in `request.h` when compiling a code with NVCC and OpenMPI:

`/home/choin/hoodoo/install/moose/libmesh/installed/include/timpi/request.h(112): error: expression must have a constant value`

It seems that the definition of `MPI_REQUEST_NULL` in OpenMPI is not considered compile-time constant by NVCC, which is:

```
OMPI_DECLSPEC extern struct ompi_predefined_request_t ompi_request_null;
#define OMPI_PREDEFINED_GLOBAL(type, global) (static_cast<type> (static_cast<void *> (&(global))))
#define MPI_REQUEST_NULL OMPI_PREDEFINED_GLOBAL(MPI_Request, ompi_request_null)
```

We might need a better way of defining `null_request` that is compatible across different MPI variants.

NamjaeChoi:
Closes #122.

Related Griffin PR : https://github.inl.gov/ncrc/griffin/pull/947

### DIFF
--- a/src/parallel/include/timpi/request.h
+++ b/src/parallel/include/timpi/request.h
@@ -108,11 +108,7 @@ public:
   // made from \p this have been cleaned up.
   void add_post_wait_work(PostWaitWork * work);
 
-#ifdef TIMPI_HAVE_MPI
-  static constexpr request null_request = MPI_REQUEST_NULL;
-#else
-  static constexpr request null_request = 0;
-#endif
+  static const request null_request;
 
 private:
   request _request;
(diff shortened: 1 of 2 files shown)


roystgnr (approved):

roystgnr:
Thanks for this.  Keep an eye on the next MOOSE libMesh update; make sure this gets in there too?

And I assume you're continuing building and testing - would you let me know when (and with what nvc++ version) you get a clean "make check"?  I'd be especially interested if you have a build with CppUnit enabled; one of our TensorTraits unit tests was failing due to a compiler bug as of last November.

PR #124, opened 2023-05-12
Sanitize and test vector<bool> specializations

roystgnr:
@NamjaeChoi - can you see if this still gives you `-fsanitize=integer` complaints?  I'll need to do a little futzing to get a clang+MPI build working myself.

### DIFF
--- a/src/parallel/include/timpi/parallel_implementation.h
+++ b/src/parallel/include/timpi/parallel_implementation.h
@@ -114,7 +114,7 @@ inline void pack_vector_bool(const std::vector<bool,A1> & vec_in,
     {
       std::size_t index = i/data_bits;
       std::size_t offset = i%data_bits;
-      vec_out[index] += (vec_in[i]?1:0) << offset;
+      vec_out[index] += (vec_in[i]?1u:0u) << offset;
     }
 }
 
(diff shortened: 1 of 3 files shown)


NamjaeChoi:
I think this removes the complaint.

NamjaeChoi:
Only the following complaint remains related with `parallel_implementation.h`:

> /Users/choin/projects/griffin/moose/scripts/../libmesh/installed/include/timpi/parallel_implementation.h:4055:3: runtime error: implicit conversion from type 'unsigned int' of value 4294967294 (32-bit, unsigned) to type 'int' changed the value to -2 (32-bit, signed)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/choin/projects/griffin/moose/scripts/../libmesh/installed/include/timpi/parallel_implementation.h:4055:3 in

roystgnr:
I'll see if I can reproduce that in a libMesh run; I think I've figured out the clang+MPI situation.  I don't suppose you can trigger it via the core MOOSE framework at most?

roystgnr:
Whoa, I seem to be able to trigger it in TIMPI alone.

roystgnr:
Is there a way to get `-fsanitize=integer` errors to be fatal?  For now I can check `run_unit_tests.sh.log`, but if I could actually stop the unit tests from passing then we could get regression coverage for this later.  The TIMPI test suite runs really fast.

I tried adding `-fno-sanitize-recover=integer` too but it didn't seem to change anything.

roystgnr:
Would you see how things work for you with these last two commits?  Those seem to fix all the fsanitize issues I can trigger at the TIMPI level alone.

NamjaeChoi:
I also confirmed that it removes all the MPI-related complaints. Thanks.

PR #129, opened 2023-11-07
Use FPE-safety flags when compiling TIMPI

roystgnr:
I don't think this is likely to ever be important, since we don't enable FPE in our test suite and we're surely never going to be doing vectorizable operations in the tiny parts of the TIMPI library that aren't header-only ... but I do want to keep ACSM in sync, and if FPE safety in TIMPI *is* ever important I'd rather we be too safe than not-safe-enough.

### DIFF
--- a/m4/timpi_set_cxx_flags.m4
+++ b/m4/timpi_set_cxx_flags.m4
@@ -23,6 +23,13 @@ AC_DEFUN([TIMPI_SET_CXX_FLAGS],
 [
   ACSM_SET_CXX_FLAGS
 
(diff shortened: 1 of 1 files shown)


roystgnr:
"Min clang", but unrelated.  Looks like we missed a recipe to update to newer containers.

roystgnr:
We're going to have some better clang containers ready soon so I'm not going to bother with a temporary update now.  I did manually test that this update backs off gracefully when a compiler is too old to support `-ftrapping-math`

PR #130, opened 2023-11-10
Update autoconf-submodule

roystgnr:
This gets us in sync with the updates needed for libMesh


roystgnr:
"Min clang" is still an environment issue

PR #131, opened 2024-06-07
Multiset (multimap, etc) support in set_union()

roystgnr:
I swear I'd *intended* to write this already, so I was as surprised as the user who needed it to find out it wasn't there.


lindsayad:
I'm curious, what was the use case?

bwspenc:
It was for something I was asking for. One of the ways to store the information I needed was to put it in a multimap. As I've thought about it more, there's a better way of doing what I wanted to do. Nonetheless, I think this would be helpful.

roystgnr:
Now I just need to recall if there was anything else on my TIMPI wishlist that I could/should knock out before I start pushing this downstream...

PR #138, opened 2025-01-03
Assert comm size is strictly greater than 0

nmnobre:
This made me doubt for a second about libMesh/libmesh#4041 but I'm pretty sure the communicator must include at least the querying process, as we shielded with `_communicator != MPI_COMM_NULL`.

### DIFF
--- a/src/parallel/src/communicator.C
+++ b/src/parallel/src/communicator.C
@@ -192,7 +192,7 @@ void Communicator::assign(const communicator & comm)
       timpi_call_mpi
         (MPI_Comm_size(_communicator, &i));
 
-      timpi_assert_greater_equal (i, 0);
+      timpi_assert_greater (i, 0);
       _size = cast_int<processor_id_type>(i);
 
       timpi_call_mpi


jwpeterson (approved):
I couldn't find anything in the [MPI docs](https://www.open-mpi.org/doc/v3.1/man3/MPI_Comm_size.3.php) that guarantees the size of a valid Communicator will always be >= 1, but it definitely seems like a reasonable assumption...

nmnobre:
> I couldn't find anything in the [MPI docs](https://www.open-mpi.org/doc/v3.1/man3/MPI_Comm_size.3.php) that guarantees the size of a valid Communicator will always be >= 1, but it definitely seems like a reasonable assumption...

They say this (which is what we're doing, so surely size must be > 0?):

"This call is often used with [MPI_Comm_rank](https://www.open-mpi.org/doc/v3.1/man3/MPI_Comm_rank.3.php) to determine the amount of concurrency available for a specific library or program. [MPI_Comm_rank](https://www.open-mpi.org/doc/v3.1/man3/MPI_Comm_rank.3.php) indicates the rank of the process that calls it in the range from 0 . . . size-1, where size is the return value of MPI_Comm_size."

roystgnr (approved):

Issue #140, opened 2025-01-31
Error calling set_union() on map with large strings


jwpeterson:
Hi @roystgnr,

We ran into an issue calling `Communicator::set_union()` on a `map<unsigned, string>` when the strings are sufficiently large. To test this, compile the code below and run on two processors while sending the output to separate files:
```
mpiexec -n 2 ./timpi_map_string_set_union_bug-opt --keep-cout --redirect-output
```

On our systems, we see the following output on two cores:

processor 0:
```
Comm size: 2
Before union: 
Map size: 147
Map first key: 1000
Map last key: 1146
After union:
Map size: 247
Map first key: 0
Map last key: 2146
ERROR! Unexpected number of map entries. Got 247, expected 294
ERROR! Unexpected first map key = 0
```

processor 1:
```
Comm size: 2
Before union: 
Map size: 147
Map first key: 2000
Map last key: 2146
After union:
Map size: 247
Map first key: 0
Map last key: 2146
ERROR! Unexpected number of map entries. Got 247, expected 294
ERROR! Unexpected first map key = 0
```

However, if you reduce the `string_size` from 32768 to 16384, the program runs correctly. Please let me know if you are at least able to reproduce this behavior.

Test code:
```
// libMesh includes
#include "libmesh/libmesh.h"
#include "libmesh/int_range.h"
#include "libmesh/communicator.h"
#include "libmesh/parallel.h"

// C++ includes
#include <iostream>
#include <string>
#include <map>

using namespace libMesh;

// Run on two cores with --keep-cout --redirect-output
int main (int argc, char ** argv)
{
  LibMeshInit init (argc, argv);

  libMesh::out << "Comm size: " << init.comm().size() << std::endl;

  std::map<unsigned int, std::string> map_of_strings;

  unsigned int string_size =
    // 1000;   // ok
    // 2048;   // ok
    // 4096;   // ok
    // 8192;   // ok
    // 16384;  // ok
    32768;  // fail
    // 163236; // fail

  // Each entry in the map will be the same test_string
  std::string test_string(/*count*/string_size, 'a');

  // Number of entries in the map on each processor
  unsigned int n_map_entries = 147;

  auto get_offset = [](unsigned int rank)
  {
    return (rank + 1) * 1000;
  };

  // Use an offset so that the map keys do not overlap
  unsigned int offset = get_offset(init.comm().rank());
  for (auto idx : make_range(n_map_entries))
    map_of_strings[offset + idx] = test_string;

[...]

PR #141, opened 2025-02-03
Pass output iterator by reference in unpack_range

roystgnr:
This fixes failures for me in John's test in #140

### DIFF
--- a/src/parallel/include/timpi/packing.h
+++ b/src/parallel/include/timpi/packing.h
@@ -1091,7 +1091,7 @@ template <typename Context, typename buffertype,
           typename OutputIter, typename T>
 inline void unpack_range (const std::vector<buffertype> & buffer,
                           Context * context,
-                          OutputIter out_iter,
+                          OutputIter & out_iter,
                           const T * /* output_type */)
 {
   // Loop through the buffer and unpack each object, returning the
--- a/src/parallel/include/timpi/packing.h
+++ b/src/parallel/include/timpi/packing.h
@@ -1086,13 +1086,17 @@ inline Iter pack_range (const Context * context,
 
 /**
  * Helper function for range unpacking
+ *
+ * We take \p out_iter by value for maximum compatibility, but we
+ * return it afterward for the use of code that needs to unpack
+ * multiple buffers to the same output iterator.
  */
 template <typename Context, typename buffertype,
           typename OutputIter, typename T>
-inline void unpack_range (const std::vector<buffertype> & buffer,
-                          Context * context,
-                          OutputIter & out_iter,
-                          const T * /* output_type */)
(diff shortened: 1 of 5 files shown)


jwpeterson:
```
In file included from ../../src/utilities/include/timpi/timpi.h:45,
                 from ../../test/parallel_unit.C:1:
../../src/parallel/include/timpi/parallel_implementation.h: In instantiation of 'void TIMPI::Communicator::allgather(const std::vector<_Tp, _Alloc>&, std::vector<std::vector<_Tp, _Alloc>, A2>&, bool) const [with T = std::__cxx11::basic_string<char>; A1 = std::allocator<std::__cxx11::basic_string<char> >; A2 = std::allocator<std::vector<std::__cxx11::basic_string<char> > >; typename std::enable_if<libMesh::Parallel::Has_buffer_type<libMesh::Parallel::Packing<T> >::value, int>::type <anonymous> = 0]':
../../test/parallel_unit.C:209:29:   required from here
../../src/parallel/include/timpi/parallel_implementation.h:2892:36: error: cannot bind non-const lvalue reference of type 'std::back_insert_iterator<std::vector<std::__cxx11::basic_string<char> > >&' to an rvalue of type 'std::back_insert_iterator<std::vector<std::__cxx11::basic_string<char> > >'
 2892 |                  std::back_inserter(recv[i]), (T*)nullptr);
      |                  ~~~~~~~~~~~~~~~~~~^~~~~~~~~
In file included from ../../src/parallel/include/timpi/communicator.h:24,
                 from ../../src/utilities/include/timpi/timpi.h:24:
../../src/parallel/include/timpi/packing.h:1101:40: note:   initializing argument 3 of 'void TIMPI::unpack_range(const std::vector<buffertype>&, Context*, OutputIter&, const T*) [with Context = void; buffertype = unsigned int; OutputIter = std::back_insert_iterator<std::vector<std::__cxx11::basic_string<char> > >; T = std::__cxx11::basic_string<char>]'
 1101 |                           OutputIter & out_iter,
      |                           ~~~~~~~~~~~~~^~~~~~~~
```
Looks like there was a reason you were passing by value for some `OutputIter` types.

jwpeterson:
Actually, it might be an easy fix if you are willing to declare a named temporary for the iterator being declared here via std::back_inserter?
```
  for (processor_id_type i=0; i != this->size(); ++i)
    unpack_range(allbuffers[i], (void *)nullptr,
                 std::back_inserter(recv[i]), (T*)nullptr);
```

roystgnr:
I don't want to potentially break user code doing similar things.  I've got an idea for a different fix.

jwpeterson:
So, it's backwards compatible, right? Do any TIMPI users actually call `unpack_range()` directly? I don't think we do it at the libmesh level, so I guess not...

jwpeterson (approved):

PR #142, opened 2025-02-03
Use copy construction on out_iter, not assignment

roystgnr:
We have users with non-assignable output iterator types; this fixes backwards compatibility for them.  I don't like doing heap allocation on a tiny little iterator here, but we only do it once for every time we allocate up to 4MB of buffers on the heap so I'm sure it won't show up in profiling.


jwpeterson (approved):
As discussed with @roystgnr, we are hopeful the `unique_ptr` stuff in this PR is a short-term backwards-compatibility change, and that once there are no non-assignable iterators left in libmesh/MOOSE, we can just get rid of it. The assumption going forward is that all iterators used in packing would have to be assignable.

PR #147, opened 2025-03-24
Fix unit tests with C++14 compilers

roystgnr:
I was about to use a C++17-only feature on purpose, then wondered how many we've been starting to use by accident...


loganharbour:
...time to require c++17? :)

roystgnr:
If it was needed for important functionality, yeah, but if it's just making unit tests a couple lines shorter then we can still support ancient compilers.  (How does the Intel C++17 support in https://en.cppreference.com/w/cpp/compiler_support/ *still* have red boxes?)

PR #149, opened 2025-03-25
timpi_pure attribute

roystgnr:
See https://github.com/libMesh/libmesh/pull/4116 for why we should have added this years ago.


src/utilities/include/timpi/timpi_macros.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef TIMPI_MACROS_H
+#define TIMPI_MACROS_H
+
+
+// Local includes
+#include "timpi/timpi_config.h"
+
+#ifdef TIMPI_HAVE_CXX20
+#  define timpi_pure [[nodiscard("Pure-function return values should not be discarded")]]
```
jwpeterson, line 28: Any reason to not call the macro `timpi_nodiscard`?

roystgnr: Mostly so we can give a more specific message.  Although "pure" is a bit questionable here (since the communication step arguably counts as a side-effect, especially if you ever don't call `verify`/`semiverify` in sync), "There's literally no reason to call this function if you're not going to use the return value" feels like it gives information to the developer, whereas other codes tend to use `[[nodiscard]]` to indicate that error handling is done via a return value that lazy programmers might ignore, that there's some resource being returned that shouldn't be leaked, or just do distinguish between in-place and returns-a-copy style mogrifiers.

I'm not actually sure when we'd want any non-"pure" nodiscard messages in our typical code, now that I think about it.  We handle error cases via exceptions, we handle resource returns via smart pointers that clean up after themselves, and if we have any returns-a-copy style methods they're probably also pure functions, since it would be weird to modify multiple different data structures but only bother to handle one as a return value.

PR #156, opened 2025-12-06
Add some more op function APIs to Communicator

lindsayad:


roystgnr (approved):
I'd strongly prefer to add some test coverage for these before merging, but beggars, choosers.

PR #158, opened 2025-12-11
Don't barrier() when we're unwinding an exception

roystgnr:
See https://github.com/libMesh/libmesh/pull/4351 for more context.


jwpeterson (approved):
I double checked that you had the `!` in there :laughing:

lindsayad (approved):

